PR #13, opened 2019-05-02
[upgrade/pack_builder] Make targetname size restriction explicit and check

oguzcanphilips:


upgrade/pack_builder/Input.hpp
```diff
@@ -6,6 +6,15 @@
 
 namespace application
 {
+    struct TargetNameTooLongException
```
rjaegers: Why a struct? You could also hide the data members and implement the what() method.

upgrade/pack_builder/Input.hpp
```diff
@@ -6,6 +6,15 @@
...
+        : std::exception
+    {
+        explicit TargetNameTooLongException(const std::string& name, int maxSize);
```
rjaegers: explicit not really necessary here.

upgrade/pack_builder/Input.hpp
oguzcanphilips: All the other exception types used in UpgradePackBuilderFacade::Build were defined as such structs with public fields. Adapted them all.

upgrade/pack_builder/Input.hpp
oguzcanphilips: Removed.

upgrade/pack_builder/SparseVector.hpp
```diff
@@ -60,6 +66,13 @@ namespace application
         : position(position)
     {}
 
+    inline char const* OverwriteException::what() const
+    {
+        std::stringstream ss;
+        ss << "Contents specified twice for memory location at address 0x" << std::hex << std::setw(8) << std::setfill('0') << position;
+        return ss.str().c_str();
```
rjaegers: You are returning a pointer to local storage here. Referring it later will lead to undefined behavior.

PR #35, opened 2021-04-30
MemoryRange: add aliases by STL convention

danielschenk:
Especially `value_type` and `const_iterator` come handy when writing tests, because they allow gmock's `ElementsAreArray` matcher to be used with a MemoryRange.


danielschenk:
Will make from recent develop instead

PR #45, opened 2021-12-21
Feature/modern cmake

rjaegers:


daantimmer:
@rjaegers I've managed to find a way to be able to generate the 'tools' while cross-compiling. Although I am not happy with the current implementation.

It looks something like this:
```cmake
    include(ExternalProject)
    ExternalProject_Add(
        tool-gens-external
        SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}"
        INSTALL_COMMAND ""
    )

    ExternalProject_Get_Property(tool-gens-external binary_dir)
    set(gen_1_tool_path ${binary_dir}/tools/gen-1/gen-1.exe)
    set(gen_2_tool_path ${binary_dir}/tools/gen-2/gen-2.exe)

    add_executable(gen-1 IMPORTED)
    add_executable(gen-2 IMPORTED)

    set_target_properties(gen-1 PROPERTIES IMPORTED_LOCATION ${gen_1_tool_path})
    set_target_properties(gen-2 PROPERTIES IMPORTED_LOCATION ${gen_2_tool_path})

    add_dependencies(gen-1 tool-gens-external)
    add_dependencies(gen-2 tool-gens-external)
```

Where gen-1 and gen-2 are some test executables that each create their own .cpp file.

This then can be used like:

```cmake
add_custom_command(
    OUTPUT sample.cpp
    COMMAND $<TARGET_FILE:gen-1>
)

add_custom_command(
    OUTPUT crud.cpp
    COMMAND $<TARGET_FILE:gen-2>
)

add_executable(tool-gen-test
    main.cpp
    sample.cpp
    crud.cpp
)
```

The issue with it right now is that I have to put the ExternalContent part in the top-level cmakelists.txt file. When adding them to a subdirectory the TARGET_FILE:gen1/2 can't be found for some reason. Even if those subdirectories are included before use.

rjaegers:
> @rjaegers I've managed to find a way to be able to generate the 'tools' while cross-compiling. Although I am not happy with the current implementation.

@daantimmer I have chosen a similar but bit different approach in the branch that I created based on your fork. Several other issues have been solved there as well. Can you take a look?

I have created export targets for almost everything. If you first do a host build, and set the CMAKE_PREFIX_PATH in your cross-build it will work as intended. Look at the CI yml's for details.

daantimmer:
@rjaegers I saw your changes indeed. I was just looking to see if there was a way to avoid having to do the additional configure+build step. But the way it has to be done is not that nice sadly.

Something else, would there be any objection to use the CMakePresets.json functionality for the build as well?

It would change this:
```yaml
      - name: Build and Test
        run: |
          cmake -B Build -DCMAKE_BUILD_TYPE=Debug
          cmake --build Build --config Debug
          pushd Build && ctest -Q -C Debug -T Test --no-compress-output --output-on-failure
```
to
```yaml
      - name: Build and Test
        run: |
          cmake --preset standalone-all
          cmake --build --preset standalone-all
          ctest --preset standalone-all
```
Which allows a developer to have the same workflow//settings as the CI. The 'Embedded Build' would change from:
```yaml
      - run: |
          cmake -B Build-Host -G Ninja -DCMAKE_BUILD_TYPE=Release -DEMIL_BUILD_TESTING=Off -DEMIL_BUILD_EXAMPLES=Off -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
          cmake --build Build-Host --config Release
          cmake -B Build -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_COMPILER_LAUNCHER=ccache -DCMAKE_PREFIX_PATH=${{ github.workspace }}/Build-Host -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-arm-gcc.cmake
          cmake --build Build --config Release
```
to:
```yaml
      - run: |
          cmake --preset standalone-tools
          cmake --build --preset standalone-tools
          cmake --preset targetlib
          cmake --build  --preset targetlib
```

daantimmer:
When compiling on windows with the Ninja (multi-config) Generator, with clang installed in path and no MSVC dev environment cmake will use the clang toolchain as a compiler.

This however will result in a non-working build:

- [ ] UartWindows.hpp fails to compile.
```
[build] B:\PROGRA~1\LLVM\bin\CLANG_~1.EXE -DEMIL_HOST_BUILD -DNOMINMAX -D_SETUPAPI_VER=_WIN32_WINNT_LONGHORN -DCMAKE_INTDIR=\"Debug\" -IE:/git/embeddedinfralib/hal/interfaces/../.. -IE:/git/embeddedinfralib/infra/util/../.. -g -Xclang -gcodeview -O0 -D_DEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrtd -std=gnu++17 -MD -MT hal/windows/CMakeFiles/hal.windows.dir/Debug/UartWindows.cpp.obj -MF hal\windows\CMakeFiles\hal.windows.dir\Debug\UartWindows.cpp.obj.d -o hal/windows/CMakeFiles/hal.windows.dir/Debug/UartWindows.cpp.obj -c E:/git/embeddedinfralib/hal/windows/UartWindows.cpp
[build] In file included from E:/git/embeddedinfralib/hal/windows/UartWindows.cpp:1:
[build] E:/git/embeddedinfralib/hal/interfaces/../..\hal/windows/UartWindows.hpp:52:13: error: default member initializer for 'baudRate' needed within definition of enclosing class 'UartWindows' outside of member functions
[build]             UartWindowsConfig() = default;
[build]             ^
[build] E:/git/embeddedinfralib/hal/interfaces/../..\hal/windows/UartWindows.hpp:66:77: note: in evaluation of exception specification for 'hal::UartWindows::UartWindowsConfig::UartWindowsConfig' needed here
[build]         UartWindows(const std::string& portName, UartWindowsConfig config = UartWindowsConfig());
[build]                                                                             ^
[build] E:/git/embeddedinfralib/hal/interfaces/../..\hal/windows/UartWindows.hpp:59:22: note: default member initializer declared here
[build]             uint32_t baudRate = CBR_115200;
[build]                      ^
[build] E:/git/embeddedinfralib/hal/interfaces/../..\hal/windows/UartWindows.hpp:52:13: error: default member initializer for 'flowControlRts' needed within definition of enclosing class 'UartWindows' outside of member functions
[build]             UartWindowsConfig() = default;
[build]             ^
[build] E:/git/embeddedinfralib/hal/interfaces/../..\hal/windows/UartWindows.hpp:60:28: note: default member initializer declared here
[build]             RtsFlowControl flowControlRts = RtsFlowControl::RtsControlDisable;
[build]                            ^
[build] E:/git/embeddedinfralib/hal/interfaces/../..\hal/windows/UartWindows.hpp:52:13: error: default member initializer for 'parity' needed within definition of enclosing class 'UartWindows' outside of member functions
[build]             UartWindowsConfig() = default;
[build]             ^
[build] E:/git/embeddedinfralib/hal/interfaces/../..\hal/windows/UartWindows.hpp:61:20: note: default member initializer declared here
[build]             Parity parity = Parity::none;
[build]                    ^
[build] 3 errors generated.
```

Some sources claim this is a compiler bug in clang as the usage in emil is correct. A solution is to remove `= default` from the `UartWindowsConfig` constructor and replacing with an empty default constructor `UartWindowsConfig(){}`
> https://stackoverflow.com/questions/43819314/default-member-initializer-needed-within-definition-of-enclosing-class-outside

I've verified this locally and this solves the build issue on Windows with clang

infra/util/transform_file_to_string_template.cpp.conf
```diff
@@ -0,0 +1,12 @@
+#include "${name}.hpp"
+#include <array>
+
+namespace ${namespace}
```
daantimmer, line 4: Would it make sense to prepend the namespace with some fixed namespace?

```c++
namespace emil::generated::${namespace}
{
    ...
}
```
?

infra/util/transform_file_to_string_template.cpp.conf
```diff
@@ -0,0 +1,12 @@
...
+{
+    namespace
```
daantimmer: Is there a specific reason to have an anonymous namespace inside another namespace?

richardapeters: I think so. The ${name}Storage variable is not public to other files, so I'd like to have it in an unnamed namespace. And I like my unnamed namespaces nested in the namespaces in which they are used.

daantimmer: Fair enough. I've just always put my anonymous namespaces at the top of the file outside any named namespace.

daantimmer:
@rjaegers please add an option to make the whole 'protobuf' feature optional. That is, allow for a flag to not `add_subdirectory` the protobuf feature. Because right now. If the embeddedinfralib is added to a project it fails out of the box because it requires to be build first (host-build, including correct path for next step) before it can be used by an embedded target. Even if the whole protobuf feature is not required.

infra/util/transform_file_to_string_template.cpp.conf
rjaegers: You have full control over the namespace, so we leave that up to the client.

PR #46, opened 2022-01-31
Feature/modern cmake

daantimmer:
Made upgrade and protobuf optional (default enabled)


CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ option(EMIL_BUILD_TESTS "Enable build of the tests" ${EMIL_DEFAULTOPT})
 option(EMIL_BUILD_EXAMPLES "Enable build of the examples" ${EMIL_DEFAULTOPT})
 cmake_dependent_option(EMIL_ENABLE_COVERAGE "Enable compiler flags for code coverage measurements" Off "EMIL_BUILD_TESTS" Off)
 
+option(EMIL_BUILD_PROTOBUF "Enable build of protobuf library" On)
+cmake_dependent_option(EMIL_BUILD_UPGRADE "Enable build of upgrade library" On "EMIL_BUILD_PROTOBUF" Off)
```
rjaegers, line 28: I would rather see upgrade.pack_keys being disabled. In that way the upgrade component could still be built and used. The consumer would only need to supply it's own key material.

protobuf/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if (NOT EMIL_BUILD_PROTOBUF)
```
rjaegers, line 1: I would rather prevent entering the subdirectory one level-up instead of this early return.

upgrade/CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if (NOT EMIL_BUILD_UPGRADE)
```
rjaegers, line 1: As already stated above: I would exclude the Protobuf key target under pack only.

CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ option(EMIL_BUILD_TESTS "Enable build of the tests" ${EMIL_DEFAULTOPT})
 option(EMIL_BUILD_EXAMPLES "Enable build of the examples" ${EMIL_DEFAULTOPT})
 cmake_dependent_option(EMIL_ENABLE_COVERAGE "Enable compiler flags for code coverage measurements" Off "EMIL_BUILD_TESTS" Off)
 
+option(EMIL_BUILD_PROTOBUF "Enable build of protobuf library" On)
```
rjaegers, line 27: I would extend the text to point-out the implications of disabling Protobuf. E.g. "when disabling echo-rpc will also be disabled". Or something along those lines.

rjaegers, line 27: I think when disabling this we should also not add the proto-stuff in external.

rjaegers (changes requested):

PR #48, opened 2022-03-09
Feature/fix for building using clang

daantimmer:
For more information w.r.t. to the default initialization issue in UartWindows.hpp, see: https://godbolt.org/z/E7qGE375K


daantimmer:
Closed because I selected the wrong target branch

PR #49, opened 2022-03-09
fix for building using clang

daantimmer:
For more information w.r.t. to the default initialization issue in UartWindows.hpp, see: https://godbolt.org/z/E7qGE375K

### DIFF
--- a/hal/windows/UartWindows.hpp
+++ b/hal/windows/UartWindows.hpp
@@ -49,7 +49,7 @@ namespace hal
                 space = SPACEPARITY
             };
 
-            UartWindowsConfig(){}
+            UartWindowsConfig() {}
 
             UartWindowsConfig(uint32_t newbaudRate, RtsFlowControl newFlowControlRts)
                 : baudRate(newbaudRate)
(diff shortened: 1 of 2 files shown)


richardapeters (approved):

hal/windows/UartWindows.hpp
```diff
@@ -49,7 +49,7 @@ namespace hal
                 space = SPACEPARITY
             };
 
-            UartWindowsConfig() = default;
+            UartWindowsConfig(){}
```
rjaegers, line 52: Can you add a space please.

```suggestion
            UartWindowsConfig() {}
```

rjaegers (changes requested):
Can you add the clang build to the GitHub Actions CI system as well? This will make sure we don't break the Clang build in the future going forward.

daantimmer:
Discussed with @rjaegers to not add a clang-based windows build to CI as the work required and the additional build time is not worth it. the macos build is already using the clang compiler. This would only add coverage for a few files only used on windows only builds.

rjaegers (approved):

PR #51, opened 2022-03-09
only EXCLUDE_FROM_ALL when building EmIL as a dependency

daantimmer:


CMakeLists.txt
```diff
@@ -56,6 +56,15 @@ else()
     set(EMIL_DEFAULTOPT Off)
 endif()
 
+# When building EmIL by itself all do not exclude any targets from all
```
rjaegers, line 59: Typo.

```suggestion
# When building EmIL by itself do not exclude any targets from all
```

rjaegers (changes requested):

services/util/test_doubles/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-add_library(services.services_test_doubles EXCLUDE_FROM_ALL STATIC)
+add_library(services.services_test_doubles ${EMIL_EXCLUDE_FROM_ALL} STATIC)
```
rjaegers, line 1: This should have been an INTERFACE library since it has no exported symbols.

```suggestion
add_library(services.services_test_doubles ${EMIL_EXCLUDE_FROM_ALL} INTERFACE)
```

PR #52, opened 2022-03-16
Always add test_helper and test_double subdirectories

daantimmer:


rjaegers (changes requested):
Please add spaces after "if" and before "(".

Please add a newline between the if-block and any subsequent block.

Example:

```
if (bla)
    do_bla_stuff()
endif()

if (bliep)
    do_bliep_stuff()
endif()
```

And yes... we should get some CMake linter/checker in our CI.

PR #82, opened 2022-07-28
feat(osal): add osal

rjaegers:
Initial implementation of an OS Abstraction Layer (OSAL) for EmbeddedInfraLib.

When an OS (threading) is required one of the supported implementation can be chosen. When one of the OS implementation is chosen EmbeddedInfraLib will expose the standard C++ threading interface implemented in terms of the chosen implementation.

I.e. if the consuming client enables ThreadX support; EmbeddedInfraLib will, amongst others, expose a <thread> header that will start a ThreadX thread when a std::thread is created.


richardapeters (approved):
It's so beautiful :heartpulse:

richardapeters (approved):

PR #98, opened 2022-10-14
Feature/apply clang format

rjaegers:


daantimmer:
Replace gtest/gmock includes with <> instead of ""

daantimmer (changes requested):
Replace gtest/gmock includes with <> instead of ""

hal/generic/TimerServiceGeneric.cpp
```diff
@@ -49,11 +50,10 @@ namespace hal
                 nextTrigger = infra::TimePoint::max();
 
                 infra::EventDispatcher::Instance().Schedule([this]()
-                {
+                    {
                     std::unique_lock<std::recursive_mutex> eventLock(mutex);
                     Progressed(Now());
-                    NextTriggerChanged();
-                });
+                    NextTriggerChanged(); });
```
richardapeters, line 56: The closing brace should be on a new line

hal/generic/TimerServiceGeneric.cpp
```diff
@@ -49,11 +50,10 @@ namespace hal
                 nextTrigger = infra::TimePoint::max();
 
                 infra::EventDispatcher::Instance().Schedule([this]()
-                {
+                    {
                     std::unique_lock<std::recursive_mutex> eventLock(mutex);
```
richardapeters, line 54: This needs an additional level of indentation

hal/interfaces/Gpio.hpp
```diff
@@ -5,14 +5,14 @@
 
 namespace hal
 {
-    enum class InterruptTrigger: uint8_t
+    enum class InterruptTrigger : uint8_t
```
richardapeters, line 8: Can clang-format put the ": uint8_t" on the next line?

infra/event/test/TestClaimableResource.cpp
```diff
@@ -123,9 +137,12 @@ TEST_F(TestClaimableResource, ClaimUrgentTakesPrecedence)
 
 TEST_F(TestClaimableResource, TwoConsecutiveReleasesBeforeReleaseOfClaimIsProcessedResultsInNoAdditionalClaimsGranted)
 {
-    claimerA.Claim([this]() { claimerA.GrantedClaim(); });
```
richardapeters, line 126: So true single-line lambdas are not possible anymore?

infra/stream/Formatter.cpp
```diff
@@ -22,17 +22,17 @@ namespace infra
...
-                break;
-            case '^':
-                align = FormatAlign::center;
-                break;
-            case '>':
-                align = FormatAlign::right;
-                break;
-            default:
-                break;
+                case '<':
```
richardapeters, line 25: I absolutely love this! Finally, a tool that understands that the contents of switches need two separate levels of indentation! Take that, Visual Studio! Woohoo!

infra/stream/StreamManipulators.hpp
```diff
@@ -5,11 +5,25 @@
 
 namespace infra
 {
-    extern const struct Text {} text;
-    extern const struct Hex {} hex;
-    extern const struct Bin {} bin;
-    extern const struct Data {} data;
-    extern const struct Endl {} endl;
+    extern const struct Text
```
richardapeters, line 8: I would wrap this section in
// clang-format off
...
// clang-format on

infra/syntax/Json.cpp
```diff
@@ -38,15 +38,21 @@ namespace infra
...
-                case '"':  return '"';
-                case '\\': return '\\';
-                case 'b': return '\b';
-                case 'f': return '\f';
-                case 'n': return '\n';
-                case 'r': return '\r';
-                case 't': return '\t';
-                case 'u':
-                {
+                case '"':
```
richardapeters, line 41: Hmm this is unfortunate. But apparently clang-format cannot put single-return-statements on the same line as a case statement, without affecting formatting other case statements.

infra/syntax/Json.cpp
```diff
@@ -848,8 +855,7 @@ namespace infra
 
         tokenizer = current;
 
-        if ((!token.Get<JsonBiggerInt>().Negative() && token.Get<JsonBiggerInt>().Value() <= std::numeric_limits<int32_t>::max())
-            || (token.Get<JsonBiggerInt>().Negative() && token.Get<JsonBiggerInt>().Value() <= static_cast<uint64_t>(-static_cast<int64_t>(std::numeric_limits<int32_t>::min()))))
+        if ((!token.Get<JsonBiggerInt>().Negative() && token.Get<JsonBiggerInt>().Value() <= std::numeric_limits<int32_t>::max()) || (token.Get<JsonBiggerInt>().Negative() && token.Get<JsonBiggerInt>().Value() <= static_cast<uint64_t>(-static_cast<int64_t>(std::numeric_limits<int32_t>::min()))))
```
richardapeters, line 858: Holy moly, a line of 297 characters! Even I would not make a line so long.

richardapeters: Perhaps this needs a // clang-format off

infra/syntax/JsonFormatter.cpp
```diff
@@ -19,14 +20,30 @@ namespace infra
         {
             switch (c)
             {
-            case '"':  stream << "\\\""; break;
```
richardapeters, line 22: We should rewrite this to have multiple fall-through case statements ending in "stream << "\\" << c;"

infra/util/BoundedString.hpp
```diff
@@ -430,7 +431,7 @@ namespace infra
 
     template<class T>
     BoundedStringBase<T>::BoundedStringBase(T* s)
-        : range(s, s + std::strlen(s))  //NOSONAR
+        : range(s, s + std::strlen(s)) // NOSONAR
```
richardapeters, line 434: Does Sonar understand // NOSONAR with the extra space?

infra/util/IntegerNormalization.hpp
```diff
@@ -9,17 +9,59 @@ namespace infra
 {
     namespace detail
     {
-        template<std::size_t N> struct SignedInteger;
-        template<> struct SignedInteger<8> { using type = int8_t; };
```
richardapeters, line 13: Another candidate for // clang-format off

services/util/MessageCommunicationCobs.hpp
```diff
@@ -14,9 +14,9 @@ namespace services
     {
     public:
         template<std::size_t MaxMessageSize>
-            using WithMaxMessageSize = infra::WithStorage<infra::WithStorage<MessageCommunicationCobs,
```
richardapeters, line 17: Hmm I really don't like this reformatting.

services/util/ConfigurationStore.cpp
```diff
@@ -206,7 +205,9 @@ namespace services
             ++operationId;
             writeRequested = false;
             writingBlob = true;
-            Serialize(*activeBlob, [this, thisId]() { inactiveBlob->Erase([this, thisId]() { BlobWriteDone(); NotifyObservers([thisId](ConfigurationStoreObserver& observer) { observer.OperationDone(thisId); }); }); });
+            Serialize(*activeBlob, [this, thisId]()
+                { inactiveBlob->Erase([this, thisId]()
+                      { BlobWriteDone(); NotifyObservers([thisId](ConfigurationStoreObserver& observer) { observer.OperationDone(thisId); }); }); });
```
richardapeters, line 210: Why does the lambda in this line get to stay on the same line, while 14 lines lower the lambda is put on a new line?

richardapeters (approved):

infra/syntax/Json.cpp
daantimmer: I would have expted clang-format to properly align (if set) too long lines

It is definitely capable of doing so.

daantimmer: But for that to work you need to set the `ColumnLimit: 0` to some vlaue that makes sense

rjaegers: Setting the ColumnLimit to something other than 0 has some side-effects. Setting it is going to be a trade-off.

It means that when you define a class with a single base it will be put on one line. And it formats lambda's differently, this might be a good thing in some cases.

See https://stackoverflow.com/questions/61082996/when-does-clang-format-break-line-of-the-inheritance-list for an example.

hal/generic/TimerServiceGeneric.cpp
rjaegers: Setting the ColumnLimit fixes this; but with the expense of having deriving classes with a single base on a single line.

```
class Derived : public Base
{}
```

hal/generic/TimerServiceGeneric.cpp
rjaegers: Setting the ColumnLimit fixes this; but with same consequence as above.

hal/interfaces/Gpio.hpp
rjaegers: Not that I know off; it even has a preference of putting things on one line. Like in the inheritance case.

infra/event/test/TestClaimableResource.cpp
rjaegers: It is when we set a ColumnLimit.

infra/util/BoundedString.hpp
rjaegers: Since we are not getting a finding on that line; apparently yes.

services/util/MessageCommunicationCobs.hpp
rjaegers: This is improved when we set ColumnLimit.

infra/syntax/Json.cpp
richardapeters: I would wait for that issue to be resolved before setting a ColumnLimit

rjaegers:
> Replace gtest/gmock includes with <> instead of ""

If I understood you correctly you would like to do this because it sorts the gtest/gmock includes differently. E.g. it sorts them together with the system headers. If this is the only issue we can fix this with the priority in IncludeCategories.

Otherwise; were would we stop? Do we also consider our other dependencies as "system headers"?

PR #99, opened 2022-10-17
Feature/warnings as error

thomas-philips:


richardapeters (approved):

richardapeters (approved):

rjaegers (changes requested):
There are still some issues left to tacke. Two builds are failing at the moment.

The OSX build has an issue with the includes from lwip. I propose to treat the lwip-2.1.1 includes as system includes.

The embedded build has issues because of the added overload. It's a difference between the std::size_t type on 32 and 64 bit architectures. We had the same issue with our stream overloads and chose a generic solution eventually. You can see that in infra/stream; it uses infra/util/IntegerNormalization.hpp.

rjaegers (approved):

infra/util/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ target_include_directories(infra.util PUBLIC
 target_compile_options(infra.util PUBLIC
     $<$<CXX_COMPILER_ID:MSVC>:
         /Zc:__cplusplus
+        /wd4068
```
rjaegers, line 12:
```suggestion
        /wd4068 # Disable unknown pragma warnings
```

infra/util/Compatibility.hpp
```diff
@@ -16,6 +16,26 @@
...
+#elif __has_cpp_attribute(gnu::fallthrough)
+#define EMIL_FALLTHROUGH [[gnu::fallthrough]]
+#elif __cplusplus && __has_cpp_attribute(clang::fallthrough)
+// Workaround for llvm.org/PR23435, since clang 3.6 and below emit a spurious
+// error when __has_cpp_attribute is given a scoped attribute in C mode.
+#define EMIL_FALLTHROUGH [[clang::fallthrough]]
+#else
+// https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough
+#define EMIL_FALLTHROUGH // Fallthrough
+// GCC recognizes fallthrough comment
```
rjaegers, line 30:
```suggestion
// gcc recognizes fall through comments as can be seen on:
// https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough
#define EMIL_FALLTHROUGH // fall through
```

PR #107, opened 2022-11-03
feat: fix sonar findings

rjaegers:


protobuf/echo/Echo.cpp
```diff
@@ -45,9 +45,8 @@ namespace services
         return echo;
     }
 
-    ServiceProxy::ServiceProxy(Echo& echo, uint32_t id, uint32_t maxMessageSize)
+    ServiceProxy::ServiceProxy(Echo& echo, EMIL_MAYBE_UNUSED uint32_t serviceId, uint32_t maxMessageSize)
         : echo(echo)
-        , serviceId(id)
```
richardapeters, line 50: this looks like a mistake. id should be used to initialize serviceId

protobuf/echo/Echo.hpp
```diff
@@ -198,7 +199,7 @@ namespace services
         : public infra::IntrusiveList<ServiceProxy>::NodeType
     {
     public:
-        ServiceProxy(Echo& echo, uint32_t id, uint32_t maxMessageSize);
+        ServiceProxy(Echo& echo, EMIL_MAYBE_UNUSED uint32_t serviceId, uint32_t maxMessageSize);
```
richardapeters, line 202: This is just the declaration; does it need an [[unused]] tag as well?

protobuf/echo/Echo.hpp
```diff
@@ -207,7 +208,6 @@ namespace services
 
     private:
         Echo& echo;
-        uint32_t serviceId;
```
richardapeters, line 210: if serviceId is not necessary, then the parameter to the constructor should be removed completely

services/network/BonjourServer.hpp
```diff
@@ -51,7 +51,6 @@ namespace services
 
         private:
             BonjourServer& server;
-            uint16_t queryId;
```
richardapeters, line 54: if queryId is not used, then it should be removed from the constructor

services/tracer/TracerWithTime.cpp
```diff
@@ -10,12 +11,10 @@ namespace services
 
     void TracerWithTime::InsertHeader()
     {
-        infra::TimePoint nowTimePoint = infra::Now();
-        time_t now = std::chrono::system_clock::to_time_t(nowTimePoint);
-        std::tm* utcTime = gmtime(&now);
```
richardapeters, line 15: Ouch I thought I had gotten rid of gmtime and friends a long time ago. Thanks!

richardapeters (approved):

protobuf/echo/Echo.hpp
rjaegers: Probably not, no. If we keep the parameter, I will remove this.

protobuf/echo/Echo.hpp
rjaegers: I have removed it completely. Including from the code generator.

services/network/BonjourServer.hpp
rjaegers: queryId is not used as a member in the class, but it is used in the contstructor to fill in the queryId of DnsRecordHeader, a temporary that is then streamed to stream.

services/tracer/TracerWithTime.cpp
rjaegers: There is still one instance left... I'll see if I can remove that as well.

PR #115, opened 2022-11-11
fix: pin version of base image to sha of ubuntu:20.04

m-beelman:
This change restores compatibility between the base image and the packages being nstalled. The list of packages is compatible with the base image of 20.04 but the modified sha256 is for ubuntu:latest.

This leads to incompatibilities that prevent the image from being built.


rjaegers:
Thanks for your PR! I noticed the same in #114. It’s time we opensource our pre-built devcontainer. I’ll close mine as duplicate.

rjaegers:
On another note: good to see what happens with the actions and repository secrets when running from a fork. This needs some attention.

richardapeters (approved):

rjaegers:
I have merged #114. And will close this one as duplicate. I’ll create an issue to look into the actions when running from a fork.

PR #121, opened 2022-11-17
Changed infra::Function to call std::abort when it has no function assigned

daantimmer:


daantimmer:
I've converted this PR to draft, pending an investigation on runtime failures when some sanitizers are enabled.

richardapeters (approved):

richardapeters (approved):

PR #122, opened 2022-11-19
feat: refactor cucumber implementation

rjaegers:


rjaegers:
Closed in favour of integrating https://github.com/philips-software/amp-cucumber-cpp-runner going forward. Cucumber support in amp-embedded-infra-lib will be marked deprecated, and eventually removed.

PR #129, opened 2022-12-07
feat: fix Sonar findings

rjaegers:


services/tracer/test/TestTracerAdapterPrintf.cpp
```diff
@@ -31,6 +31,18 @@ class TracerAdapterPrintfTest
...
 
+TEST_F(TracerAdapterPrintfTest, print_newline_to_tracer)
+{
+    Print("\n");
+    EXPECT_EQ("\r\n", stream.Storage());
+}
+
+TEST_F(TracerAdapterPrintfTest, null_character_prints_nothing)
+{
+    Print("\0");
```
richardapeters, line 42: You can't put a 0 character inside a null-terminated string. This test case is exactly the same as Print ("");

richardapeters (approved):

services/tracer/test/TestTracerAdapterPrintf.cpp
rjaegers: Doh! I have removed the test.

PR #131, opened 2022-12-08
feat: generalize examples

rjaegers:
- added static library that abstracts networking differences between os's
- used library in examples
- fixed several issues in BSD network implementation


services/network_instantiations/NetworkAdapter.hpp
```diff
@@ -0,0 +1,37 @@
+#ifndef SERVICES_NETWORK_INSTANTIATIONS_HPP
+#define SERVICES_NETWORK_INSTANTIATIONS_HPP
+
+#include "services/network/Connection.hpp"
```
richardapeters, line 4: These includes are now not necessary anymore

PR #136, opened 2022-12-12
feat: hal unix

rjaegers:
OSX implementation inspired by: https://developer.apple.com/library/archive/samplecode/SerialPortSample/Introduction/Intro.html#//apple_ref/doc/uid/DTS10000454-Intro-DontLinkElementID_2


BarisTanyeri (approved):
Unix FTW!

PR #141, opened 2022-12-19
feat: add rpc example

rjaegers:


examples/rpc/Server.cpp
```diff
@@ -0,0 +1,65 @@
...
+        if (connection.Allocatable())
+            createdObserver(connection.Emplace());
+    }
+
+private:
+    infra::SharedPtr<void> listener;
+    infra::SharedOptional<EchoConnection> connection;
+};
+
+int main(int argc, const char* argv[], const char* env[])
```
daantimmer, line 56: TIL that 'env' is a thing.

From cppreference:
> A very common implementation-defined form of main() has a third argument (in addition to argc and argv), of type char**, pointing at [an array of pointers to the execution environment variables](https://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html).

Never heard of this, _ever_, before.

daantimmer (approved):

examples/rpc/Server.cpp
rjaegers: It's a bit of legacy in our examples. We used to have an environment variable pointing to a configuration file. We used the 'env' argument to find it. Now it's not used like that anymore but the 'env' in the main signature stayed.

PR #147, opened 2022-12-20
feat: add release-please action

rjaegers:
Use the release-please action to automatically create release PRs for every change that lands in the main branch.


.release-please-manifest.json
```diff
@@ -0,0 +1,3 @@
+{
```
daantimmer, line 1: Is it indeed required to duplicate the version number? No way to extract it from cmake? Or inject it via tags?

rjaegers: I wouldn't say duplicate. This version number is now leading, it's the one-and-only ground-truth. The version number in the CMakeLists.txt and, for example, the antora documentation will be updated as part of this action.

When the PR that this action generates is merged back to the main branch the tags and releases are created automatically.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,64 @@
...
+        id: token
+        with:
+          app_id: ${{ secrets.FOREST_RELEASER_APP_ID }}
+          app_base64_private_key: ${{ secrets.FOREST_RELEASER_APP_PRIVATE_KEY_BASE64 }}
+          auth_type: installation
+      - uses: google-github-actions/release-please-action@af1990728f970ad54c88e130c02a1521988127a2 # v3.7.0
+        id: release
+        with:
+          command: manifest
+          #token: ${{ steps.token.outputs.token }}
```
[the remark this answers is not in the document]
rjaegers: I would like to keep this in. When the app that generates the token has the correct scopes, we need to switch from the current PAT to this token again. When that happens also the bootstrap-sha needs to be removed.

daantimmer (approved):

PR #166, opened 2023-01-11
feat: add cmake/emil_antora.cmake with command generate_antora_docs_target

richardapeters:


cmake/emil_antora.cmake
```diff
@@ -0,0 +1,6 @@
+function(generate_antora_docs_target playbook)
+    add_custom_target(generate_antora_docs
+        COMMAND antora --cache-dir=.cache --stacktrace --fetch --generator=antora-site-generator-lunr ${playbook}
```
rjaegers, line 3: This will only work when:

- You have a local (working) antora installation
- We add antora to the devcontainer

I think it will be better to use the approach that we also use for running the linter locally. Start a Docker container with Antora in it to do the generation work. This will scale better without having a direct impact on the size of our devcontainer.

rjaegers (changes requested):
See the inline comment about where to run the command.

cmake/emil_antora.cmake
richardapeters: Yes, but we agreed that that would be a second step which you would implement later on :)

rjaegers: 🤡

rjaegers (approved):

PR #171, opened 2023-01-13
feat: apply clang-format to all files

richardapeters:


CMakeLists.txt
```diff
@@ -188,3 +188,7 @@ if (EMIL_GENERATE_PACKAGE_CONFIG)
 
     include(CPack)
 endif()
+
+if (EMIL_STANDALONE)
+    emil_clangformat_directory(${CMAKE_CURRENT_LIST_DIR} emil)
+endif()
```
rjaegers, line 194: Create this target unconditionally. There is no real need to keep it out when consuming emil as a submodule or via fetchcontent.

richardapeters: Done

PR #182, opened 2023-01-24
build: remove seanmiddleditch/gha-setup-ninja

rjaegers:
The seanmiddleditch/gha-setup-ninja action has not yet moved from node12 to node16 as is recommended by GitHub (https://github.com/seanmiddleditch/gha-setup-ninja/issues/16). This gives unnecessary warnings in our build summaries.

Replace the usage of the action by installing ninja via the package manager.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
         uses: carlosperate/arm-none-eabi-gcc-action@48db4484a55750df7a0ccca63347fcdea6534d78
         with:
           release: ${{ matrix.gcc }}
-      - uses: seanmiddleditch/gha-setup-ninja@16b940825621068d98711680b6c3ff92201f8fc0
+      - run: sudo apt-get update && sudo apt-get install ninja-build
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         uses: carlosperate/arm-none-eabi-gcc-action@48db4484a55750df7a0ccca63347fcdea6534d78
         with:
           release: ${{ matrix.gcc }}
-      - uses: seanmiddleditch/gha-setup-ninja@16b940825621068d98711680b6c3ff92201f8fc0
+      - run: sudo apt-get update && sudo apt-get install ninja-build
```
EkelmansPh, line 65: This will always get the latest version, right? Can we expect here that the latest version is always compatible with our environment?

EkelmansPh (approved):

.github/workflows/ci.yml
rjaegers: The ninja package is stable enough to trust that it will. Although we can eventually switch to a devcontainer for all our CI jobs as well. This would pin all dependencies.

PR #224, opened 2023-03-10
chore: create universal binaries for OS X (10.15 and higher)

rjaegers:
Build a universal binaries supporting both x86_64 and arm64 architectures.
Set the minimum supported target version to 10.15.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.24)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 
-if (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+if (APPLE)
     set(CMAKE_OSX_ARCHITECTURES "arm64;x86_64" CACHE STRING "OS X target architectures")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,6 +1,11 @@
 cmake_minimum_required(VERSION 3.24)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 
+if (CMAKE_SYSTME_NAME STREQUAL "Darwin")
```
richardapeters, line 4:
```suggestion
if (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
```

CMakeLists.txt
```diff
@@ -1,6 +1,11 @@
...
+    set(CMAKE_OSX_ARCHITECTURES "arm64;x86_64" CACHE STRING "OS X target architectures")
+    set(CMAKE_OSX_DEPLOYMENT_TARGET "10.14" CACHE STRING "Minimum OS X deployment version")
```
richardapeters, line 6: This version says 10.14 but the feature title says 10.15. Which one is it?

richardapeters (approved):

CMakeLists.txt
rjaegers: 🤡

CMakeLists.txt
rjaegers: 🤡

PR #225, opened 2023-03-11
fix: decouple host and embedded builds

rjaegers:
# Description

When building for embedded targets it is no longer needed to first do a host build of amp-embedded-infra-lib.

# Details

When using the Echo (RPC) framework based on Google Protocol Buffers, the compiler and compiler plug-ins are fetched from GitHub releases on-demand when building for embedded targets.

In case of a host build the built compiler plug-ins are used, in case of an embedded build the plug-ins are either consumed from a local amp-embedded-infra-lib install or downloaded via FetchContent.

## How to use installed echo plug-ins

When wanting to use locally built echo plug-ins in either amp-embedded-infra-lib or a consuming project (i.e. a project _using_ amp-embedded-infra-lib) when doing an embedded build; it is necessary to:

- Set the CMAKE_INSTALL_PREFIX to a known location (amp-embedded-infra-lib uses CMakePresets.json to set it to ${sourceDir}/install)
- Add the CMAKE_INSTALL_PREFIX to the CMAKE_PREFIX_PATH (e.g. list(APPEND CMAKE_PREFIX_PATH "${CMAKE_INSTALL_PREFIX}"), this is necessary since the prefix paths are empty for a cross-compile.
- Build and install the host configuration (resulting in an `install` directory)
- Build the embedded configuration (check the configure step for a log line that the installed echo plug-ins are used)


CMakeLists.txt
```diff
@@ -69,31 +69,20 @@ else()
...
-if (CMAKE_CROSSCOMPILING)
-    set(EMIL_PACKAGE_CONFIG_IMPORT_NAMESPACE "${PROJECT_NAME}::" CACHE STRING "Namespace used to reference imported targets")
-else()
-    set(EMIL_PACKAGE_CONFIG_IMPORT_NAMESPACE "" CACHE STRING "Namespace used to reference imported targets")
-endif()
 
 option(EMIL_INCLUDE_MBEDTLS "Include MbedTLS as part of EmIL" On)
 set(EMIL_EXTERNAL_LWIP_TARGET "" CACHE STRING "Specify an external LWIP target")
 
+if (EMIL_COMPILE_WARNING_AS_ERROR)
```
richardapeters, line 82: Why do we need this variable? Can't the CMake variable just be set in the CMakePresets by a user that wants to compile with warnings as errors?

external/protoc/CMakeLists.txt
```diff
@@ -21,17 +21,13 @@ function(emil_fetch_protocol_buffer_compiler)
     FetchContent_MakeAvailable(protoc)
 
     if (CMAKE_HOST_WIN32)
-        set(EMIL_PROTOC_COMPILER_BINARY "${protoc_SOURCE_DIR}/bin/protoc.exe" CACHE INTERNAL "")
-    else()
-        set(EMIL_PROTOC_COMPILER_BINARY "${protoc_SOURCE_DIR}/bin/protoc" CACHE INTERNAL "")
+        set(host_executable_postfix ".exe")
```
richardapeters, line 24: Can CMAKE_EXECUTABLE_SUFFIX be used instead of this?

protobuf/echo/protocol_buffer_echo.cmake
```diff
@@ -1,8 +1,34 @@
-function(protocol_buffer_echo_generator target input)
-    if (CMAKE_CROSSCOMPILING)
-        find_package(emil COMPONENTS Infra Hal Crypto Services Protobuf REQUIRED)
+function(emil_fetch_echo_plugins)
+    set(emil_version "3.0.0") # x-release-please-version
+
+    if (CMAKE_HOST_WIN32)
+        FetchContent_Declare(echoplugins
+            URL https://github.com/philips-software/amp-embedded-infra-lib/releases/download/v${emil_version}/emil-${emil_version}-win64.zip
```
richardapeters, line 6: Can you extract a function which gives the download suffix for the current platform?

protobuf/echo/protocol_buffer_echo.cmake
```diff
@@ -1,8 +1,34 @@
...
+        message(FATAL_ERROR "No suitable echo plugin found for ${CMAKE_HOST_SYSTEM_NAME} (${CMAKE_HOST_SYSTEM_PROCESSOR})")
+    endif()
+
+    FetchContent_MakeAvailable(echoplugins)
+
+    if (CMAKE_HOST_WIN32)
+        set(host_executable_postfix ".exe")
     endif()
 
+    set(EMIL_ECHO_CPP_COMPILER_BINARY "${echoplugins_SOURCE_DIR}/bin/protobuf.protoc_echo_plugin${host_executable_postfix}" CACHE INTERNAL "")
```
richardapeters, line 26: Perhaps this is a nice opportunity to use add_executable( ... IMPORTED GLOBAL)?

richardapeters: It would probably remove the duplication in the custom command below

external/protoc/CMakeLists.txt
rjaegers: No, because that is for the target not the host. So if you cross-compile it can be set to .elf.

protobuf/echo/protocol_buffer_echo.cmake
daantimmer: I would even perhaps opt for allowing custom urls?
`EMIL_OPTIONAL_BINARY_RELEASE_URL`
If set it uses this url, else the default download location? (Thinking about how some projects would react to seeing a dependency om some open source tooling/hosting). I would assume they would rather use a local hosting solution instead.

external/protoc/CMakeLists.txt
richardapeters: Right, of course!

protobuf/echo/protocol_buffer_echo.cmake
rjaegers: I don't think that will be necessary. We use a fully transparent and auditable release process. Only the release bot creates (signed) releases.

Next to this dependency we already have other dependencies to open-source projects.

I do still think that promoting the plug-in binary variables to internal cached variables will solve most of the issues.

richardapeters (approved):

.gitignore
```diff
@@ -1,6 +1,7 @@
 .cache/
 .vs/
 [Bb]uild/
```
rjaegers, line 3:
```suggestion
build/
```

PR #230, opened 2023-03-19
feat(echo_console): add support for ECHO over websockets over http

richardapeters:
… http


protobuf/echo_console/Main.cpp
```diff
@@ -279,8 +348,10 @@ int main(int argc, char* argv[], const char* env[])
                 uart.Emplace(get(target));
                 consoleClientUart.Emplace(console, *uart);
             }
+            else if (services::SchemeFromUrl(get(target)) == "http")
```
rjaegers, line 351: The WebSocket protocol defines a ws:// prefix to indicate a WebSocket connection. I think we should use that here. Or maybe both?

PR #265, opened 2023-04-20
build: add clang x86_64 Windows toolchain file

rjaegers:
This PR adds support for cross-compilation from non-Windows hosts to Windows hosts.
It assumes the Windows SDK and CRT are installed at a path specified by WINDOWS_SDK_ROOT.

The recommended way of installing the SDK and CRT is by using xwin (https://github.com/Jake-Shadle/xwin):
`$ xwin --accept-license splat --preserve-ms-arch-notation`

xwin patches the SDK and CRT for case-sensitive filesystems; the --presere-ms-arch-notation is
necessary to make the /winsdkdir and /vctoolsdir options of clang work.

### DIFF
--- a/cmake/toolchain-x86_64-pc-windows-msvc.cmake
+++ b/cmake/toolchain-x86_64-pc-windows-msvc.cmake
@@ -4,7 +4,7 @@
 # The recommended way of installing the SDK and CRT is by using xwin (https://github.com/Jake-Shadle/xwin):
 # $ xwin --accept-license splat --preserve-ms-arch-notation
 #
-# xwin patches the SDK and CRT for case-sensitive filesystems; the --presere-ms-arch-notation is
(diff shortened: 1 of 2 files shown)


cmake/toolchain-x86_64-pc-windows-msvc.cmake
```diff
@@ -0,0 +1,22 @@
+# This toolchain file enables cross-compilation from non-Windows hosts to Windows hosts.
+# It assumes the Windows SDK and CRT are installed at a path specified by WINDOWS_SDK_ROOT.
+#
+# The recommended way of installing the SDK and CRT is by using xwin (https://github.com/Jake-Shadle/xwin):
+# $ xwin --accept-license splat --preserve-ms-arch-notation
+#
+# xwin patches the SDK and CRT for case-sensitive filesystems; the --presere-ms-arch-notation is
```
richardapeters, line 7: --presere -> --preserve

PR #269, opened 2023-04-23
build: add emil_build_for macro

richardapeters:


cmake/emil_build_for.cmake
```diff
@@ -0,0 +1,87 @@
+# 'emil_build_for' uses the EXCLUDE_FROM_ALL property on targets to specify whether the
+# target should be built for a certain combination of properties.
+#
+# These properties are supported:
+#
+# HOST (values: All, Windows, Linux, Darwin): Build the target when building for any all of the host OSes
+# TARGET_MCU_VENDOR: Build the target when building for this vendor
+# TARGET_MCU_FAMILY: Build the target when building for this MCU family
+# TARGET_MCU: Build the target when building for this MCU
+# TARGET_MCU VARIANT: Build the target when building for this MCU variant
```
rjaegers, line 10:
```suggestion
# TARGET_MCU_VARIANT: Build the target when building for this MCU variant
```

cmake/emil_build_for.cmake
```diff
@@ -0,0 +1,87 @@
...
+# BOOL: Build the target when this boolean value evaluates to true
+#
+# PREREQUISITE_BOOL: Build the target only when this boolean value evaluates to true
+# PREREQUISITE_CONFIG: Build the target only when building for this configuration
+#
+# The HOST, TARGET_*, and BOOL properties are ORed together. Its result is ANDed with the PREREQUISITE_* properties.
+# Specifying PREREQUISITE_CONFIG results in a generator expression being ued for EXCLUDE_FROM_ALL
```
rjaegers, line 17:
```suggestion
# Specifying PREREQUISITE_CONFIG results in a generator expression being used for EXCLUDE_FROM_ALL
```

cmake/emil_build_for.cmake
```diff
@@ -0,0 +1,87 @@
+# 'emil_build_for' uses the EXCLUDE_FROM_ALL property on targets to specify whether the
+# target should be built for a certain combination of properties.
+#
+# These properties are supported:
```
rjaegers, line 4: Should we describe that all properties are optional?

I.e. that you could also use the function like this:
`emil_buid_for(target HOST Linux)`

rjaegers (approved):

cmake/emil_build_for.cmake
richardapeters: I'll add that as an  example

PR #278, opened 2023-04-29
refactor: fix lifetime issue in infra/syntax/test/TestProtoParser

richardapeters:


infra/util/test/TestAllocatorFixedSpace.cpp
```diff
@@ -13,6 +13,7 @@ class FixedSpaceAllocatorTest
 
 TEST_F(FixedSpaceAllocatorTest, allocate_one_object)
 {
+    EXPECT_TRUE(false);
```
rjaegers, line 16: Is this a test to see if we pay attention 😅.

rjaegers (changes requested):

infra/util/test/TestAllocatorFixedSpace.cpp
richardapeters: :sweat_smile: No I wanted to see the workflow storing the results on a failed run

daantimmer (approved):
Approved as far as I can judge from reviewing from my phone

rjaegers (approved):

PR #279, opened 2023-05-03
feat: merge upstream changes

richardapeters:


infra/util/test_helper/MockCallback.hpp
```diff
@@ -106,6 +106,26 @@ namespace infra
             };
         }
     };
+
+    template<class T, class P1, class P2, class P3, std::size_t ExtraSize>
+    class VerifyingFunctionMock<T(P1, P2, P3), ExtraSize>
```
daantimmer, line 111: Why isn't this a variadic template parameter for P1 P2 and P3?

daantimmer (approved):
There is only one minor comment, but didn't need to be resolved for this PR

infra/util/test_helper/MockCallback.hpp
richardapeters: I'll look into this later

PR #290, opened 2023-05-19
fix: ensure forwarding constructors have constraints

richardapeters:
ensure forwarding constructors have constraints that prevent copying / moving objects of the same type

### DIFF
--- a/infra/stream/InputStream.hpp
+++ b/infra/stream/InputStream.hpp
@@ -63,9 +63,9 @@ namespace infra
         StreamErrorPolicy& errorPolicy;
     };
 
-    template<class Parent, class TheReader>
+    template<class Parent, class ReaderType>
     class InputStreamWithReader
-        : private detail::StorageHolder<TheReader, InputStreamWithReader<Parent, TheReader>>
+        : private detail::StorageHolder<ReaderType, InputStreamWithReader<Parent, ReaderType>>
         , public Parent
     {
     public:
@@ -82,7 +82,7 @@ namespace infra
         InputStreamWithReader& operator=(const InputStreamWithReader& other) = delete;
         ~InputStreamWithReader() = default;
 
-        TheReader& Reader();
+        ReaderType& Reader();
 
     private:
         StreamErrorPolicy errorPolicy;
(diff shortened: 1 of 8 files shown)


infra/stream/InputStream.hpp
```diff
@@ -63,13 +63,53 @@ namespace infra
         StreamErrorPolicy& errorPolicy;
     };
 
+    template<class Parent, class TheReader>
```
rjaegers, line 66: Can we find a better name for TheReader? E.g. ReaderType?

infra/stream/OutputStream.hpp
```diff
@@ -82,13 +82,53 @@ namespace infra
         StreamErrorPolicy& errorPolicy;
     };
 
+    template<class Parent, class TheWriter>
```
rjaegers, line 85: Same comment for TheWriter, WriterType?

rjaegers (approved):
Can you check the remaining six critical code-smells? It seems they are easy to fix.

infra/stream/InputStream.hpp
```diff
@@ -63,13 +63,53 @@ namespace infra
...
+    class InputStreamWithReader
+        : private detail::StorageHolder<TheReader, InputStreamWithReader<Parent, TheReader>>
+        , public Parent
+    {
+    public:
+        InputStreamWithReader();
+        template<class Arg>
+        explicit InputStreamWithReader(Arg&& arg, std::enable_if_t<!std::is_same_v<InputStreamWithReader, std::remove_cv_t<std::remove_reference_t<Arg>>>, std::nullptr_t> = nullptr);
```
rjaegers, line 74: I created a wrapper in util/Compatibility.hpp for enable_if_t. Do we still want to provide C++11 compatibility or do we drop that completely?

rjaegers (commented):
Should this be a breaking change? Since it changes the interface of  the stream readers and writers?

richardapeters:
> Should this be a breaking change? Since it changes the interface of the stream readers and writers?

Technically it breaks the interface, but not in a way that should matter. We prevent the incorrect constructor being called

infra/stream/InputStream.hpp
richardapeters: Sounds good, I'll apply it

infra/stream/OutputStream.hpp
richardapeters: Same answer, sounds good, I'll apply it :)

PR #291, opened 2023-05-19
fix: correct order of destruction in EchoForwarderToSerial.

EkelmansPh:
Remove all observers of EchoForwarder on destruction of EchoForwarderToSerial before destroying subject EchoOnSerialCommunication.


protobuf/echo/EchoInstantiation.hpp
```diff
@@ -68,6 +73,11 @@ namespace main_
             , to(toSerial)
         {}
 
+        ~EchoForwarderToSerial()
+        {
+            this->RemoveAllForwarders();
```
rjaegers, line 78: Why not just a forwarders.clear() here?

rjaegers (changes requested):

protobuf/echo/EchoInstantiation.hpp
richardapeters: Indeed; I don't see why someone outside of a destruction situation would want to remove all forwarders in one go

EkelmansPh: Because I didn't want to increase the scope of the forwarders object outside of the EchoForwarder class. Is it an idea to  make the RemoveAllForwarders method protected instead?

rjaegers: Aah, I missed a bit of context in the diff.

In that case, why not add a destructor to the base class to clean-up its resources.

EkelmansPh: After discussion, I have now made the echoForwarder an object of EchoForwarderToSerial instead of inheriting from it. This resolves the problem with order of destruction in a cleaner way.

rjaegers (approved):

PR #303, opened 2023-05-25
refactor: remove virtual where final or override are used

rjaegers:
As per: https://github.com/isocpp/CppCoreGuidelines/blob/036324/CppCoreGuidelines.md#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final

Or per Sonar: https://rules.sonarsource.com/cpp/RSPEC-3471


oguzcanoguz:
What is the rationale here?
Is virtual redundant in those cases?

rjaegers:
> What is the rationale here? Is virtual redundant in those cases?

Yes, virtual is redundant in those cases.

See: https://github.com/isocpp/CppCoreGuidelines/blob/036324/CppCoreGuidelines.md#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final

richardapeters (approved):

PR #304, opened 2023-05-25
fix: usable memory loss in HttpClientBasic::Path

BarisTanyeri:
This change is a one liner.

When Path is extracted, resulting BoundedString should have a internal range of size ;
 - maxSize - hostUrlSize but previously it was maxSize - hostUrlSize*2

What was the issue?
- String range was also trimmed at the end, the size of the host url.

### DIFF
--- a/services/network/HttpClientBasic.cpp
+++ b/services/network/HttpClientBasic.cpp
@@ -73,7 +73,7 @@ namespace services
     {
         auto path = services::PathFromUrl(url);
         auto offset = path.empty() ? url.size() : path.begin() - url.begin();
-        return infra::BoundedString(infra::MakeRange(url.begin() + offset, url.begin() + url.max_size() - offset), path.size());
+        return infra::BoundedString(infra::MakeRange(url.begin() + offset, url.begin() + url.max_size()), path.size());
     }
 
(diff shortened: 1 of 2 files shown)


services/network/HttpClientBasic.cpp
```diff
@@ -73,7 +73,7 @@ namespace services
     {
         auto path = services::PathFromUrl(url);
         auto offset = path.empty() ? url.size() : path.begin() - url.begin();
-        return infra::BoundedString(infra::MakeRange(url.begin() + offset, url.begin() + url.max_size() - offset), path.size());
+        return infra::BoundedString(infra::MakeRange(url.begin() + offset, url.begin() + url.max_size()), path.size());
```
richardapeters, line 76: Good find! Can you also add one or a few microtests on the HttpClientBasic::Path() function that verifies the correct resolution of the bug?

richardapeters (approved):

services/network/HttpClientBasic.cpp
BarisTanyeri: I have added 2 tests
- one that checks whether Path() returns the expected value
- one that checks whether the returned range has the expected size

PR #307, opened 2023-05-26
fix: Line termination mismatch in hex files

andjordan:
A .hex file might have line termination with LF (Unix style) or CRLF (Windows style)
Given a .hex with CRLF termination and a pack_builder running on Linux, it would fail to convert the .hex file with Record Too Long exception

### DIFF
--- a/upgrade/pack_builder/BinaryObject.cpp
+++ b/upgrade/pack_builder/BinaryObject.cpp
@@ -2,6 +2,14 @@
 #include "infra/stream/StdStringInputStream.hpp"
 #include <algorithm>
 
+namespace
+{
+    bool RemainderHasNoData(const infra::StdStringInputStream& stream)
+    {
+        return !std::all_of(stream.PeekContiguousRange().begin(), stream.PeekContiguousRange().end(), [](auto c) { return std::isspace(c); });
+    }
+}
+
 namespace application
 {
     LineException::LineException(const std::string& prependMessage, const std::string& file, int line)
@@ -39,7 +47,7 @@ namespace application
         this->offset = offset;
 
         int lineNumber = 0;
-        for (auto line : data)
+        for (const auto& line : data)
         {
             ++lineNumber;
             if (!line.empty())
@@ -148,7 +156,7 @@ namespace application
             memory.Insert(lineContents.data[i], linearAddress + offset + lineContents.address + i);
     }
 
-    BinaryObject::LineContents::LineContents(std::string line, const std::string& fileName, int lineNumber)
+    BinaryObject::LineContents::LineContents(const std::string& line, const std::string& fileName, int lineNumber)
     {
         infra::StdStringInputStream stream(line, infra::softFail);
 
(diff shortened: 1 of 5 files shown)


upgrade/pack_builder/BinaryObject.cpp
```diff
@@ -2,6 +2,15 @@
 #include "infra/stream/StdStringInputStream.hpp"
 #include <algorithm>
 
+namespace
+{
+    bool RemainderHasNoData(const infra::StdStringInputStream& stream)
```
richardapeters, line 7: The name of this function talks about remainder and data, but the implementation is of a much lower level. I also think the name says the inverse of what it actually returns. Please consider remaining to something like StreamHasNonSpaces

upgrade/pack_builder/BinaryObject.cpp
```diff
@@ -2,6 +2,15 @@
...
+    {
+        return !std::all_of(stream.PeekContiguousRange().begin(), stream.PeekContiguousRange().end(), [](auto c)
```
richardapeters, line 9: If you want to examine the rest of the stream, you should not make the assumption that the remainder is contiguous. Instead, repeatedly invoke PeekContiguousRange, passing a start at each invocation to make sure you examine everything

upgrade/pack_builder/test/TestBinaryObject.cpp
```diff
@@ -127,6 +127,15 @@ TEST(BinaryObjectTest, Hex_TooLongRecord)
     EXPECT_THROW(object.AddHex(std::vector<std::string>{ ":0100000001fe0" }, 0, "file"), application::RecordTooLongException);
 }
 
+TEST(BinaryObjectTest, Hex_TrailingLineTerminator)
+{
+    application::BinaryObject object;
+    object.AddHex(std::vector<std::string>{ ":00000001FF\r" }, 0, "file");
+
+    application::SparseVector<uint8_t> expectedMemory;
+    EXPECT_EQ(expectedMemory, object.Memory());
```
richardapeters, line 136: Does this work?
```suggestion
    EXPECT_EQ({}, object.Memory());
```

upgrade/pack_builder/BinaryObject.cpp
andjordan: oops, thanks :)

upgrade/pack_builder/BinaryObject.cpp
```diff
@@ -2,6 +2,15 @@
...
+            { return std::isspace(c); });
```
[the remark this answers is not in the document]
andjordan: I think it makes sense to change to `\r` as it is locale dependent, however this is host only

upgrade/pack_builder/test/TestBinaryObject.cpp
andjordan: No, but `EXPECT_EQ(application::SparseVector<uint8_t>{}, object.Memory());` does

upgrade/pack_builder/BinaryObject.cpp
```diff
@@ -178,7 +193,8 @@ namespace application
             throw RecordTooShortException(fileName, lineNumber);
         if (sum != 0)
             throw IncorrectCrcException(fileName, lineNumber);
-        if (!stream.Empty())
+
+        if (!stream.Empty() && StreamHasNonSpaces(stream))
```
richardapeters, line 197:
```suggestion
        if (!stream.Empty() && StringHasNonSpaces(line.substr(line.size() - stream.Available())))
```

PR #331, opened 2023-06-21
feat: support payloads bigger than the MaxSendStreamSize

richardapeters:


services/network/Http.hpp
```diff
@@ -125,6 +126,7 @@ namespace services
         HttpHeader hostHeader;
         const HttpHeaders headers;
         bool chunked = false;
+        bool sentHeader = false;
```
daantimmer, line 129:
```suggestion
        bool chunked{false};
        bool sentHeader{false};
```

services/network/HttpClientImpl.cpp
```diff
@@ -360,11 +360,21 @@ namespace services
     void HttpClientImpl::SendingStateRequest::SendStreamAvailable(infra::SharedPtr<infra::StreamWriter>&& writer)
     {
         infra::TextOutputStream::WithErrorPolicy stream(*writer);
-        client.request->Write(stream);
-        client.request = infra::none;
-        writer = nullptr;
+        auto amount = client.request->Write(stream);
+        client.request->Consume(amount);
```
daantimmer, line 364:
```suggestion
        client.request->Consume(client.request->Write(stream));
```

services/network/Http.cpp
```diff
@@ -100,25 +100,40 @@ namespace services
...
+    std::size_t HttpRequestFormatter::Write(infra::TextOutputStream stream) const
     {
-        stream << verb << sp << requestTarget << sp << httpVersion << crlf;
+        if (!sentHeader)
+        {
+            stream << verb << sp << requestTarget << sp << httpVersion << crlf;
 
-        for (auto&& header : headers)
-            stream << header << crlf;
+            for (auto&& header : headers)
```
daantimmer, line 115:
```suggestion
            for (const auto& header : headers)
```

services/network/test_doubles/ConnectionStub.hpp
```diff
@@ -30,6 +30,7 @@ namespace services
         void SimulateDataReceived(infra::ConstByteRange data);
 
         std::vector<uint8_t> sentData;
+        std::size_t maxSendStreamSize = 1024;
```
daantimmer, line 33:
```suggestion
        const std::size_t maxSendStreamSize{ 1024 };
```

richardapeters: The intention was that tests can change this parameter

PR #335, opened 2023-06-26
feat: add order to pack upg supported targets

magi-arun:
Add order option to add the supported targets in the specified order for pack builder. Pack builder uses the order option when the 
 targets in the upgrade pack are to be added in particular sequence.  Multiple targets with similar sequence can be of the same order.


rjaegers:
I would like to propose an alternative solution. Instead of adding the 'order' property per-target, why not add an overall option that enforces the order of specification?

This would enable the library to throw a meaningful exception when the order is broken.

`application::SupportedTargets::Create().InOrder().AddHex("hextarget").AddBin("bintarget");`

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.cpp
```diff
@@ -20,6 +20,14 @@ namespace main_
         {}
     };
 
+    struct IncorrectOrderOfTargetException
+        : std::runtime_error
+    {
+        explicit IncorrectOrderOfTargetException(const std::string& target)
+            : std::runtime_error("Incorrect order of target : " + target)
```
rjaegers, line 27:
```suggestion
            : std::runtime_error("Incorrect order of target: " + target)
```

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.cpp
```diff
@@ -51,12 +59,39 @@ namespace main_
...
+        }
+        return 0;
+    }
+
     std::vector<std::unique_ptr<application::Input>> UpgradePackBuilderFacade::CreateInputs(const application::SupportedTargets& supportedTargets, const TargetAndFiles& requestedTargets, application::InputFactory& factory)
     {
         std::vector<std::unique_ptr<application::Input>> inputs;
 
+        uint8_t currentOrderOfTarget = 1;
+        auto orderedTargets = supportedTargets.OrderOfTargets();
```
rjaegers, line 81:
```suggestion
        const auto& orderedTargets = supportedTargets.OrderOfTargets();
```

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.hpp
```diff
@@ -36,6 +36,7 @@ namespace main_
         virtual void PostBuilder(application::UpgradePackBuilder& builder, application::ImageSigner& signer, const BuildOptions& buildOptions);
 
     private:
+        uint8_t GetOrder(const std::string& targetName, const std::vector<std::vector<std::string>>& orderedTargets);
```
rjaegers, line 39:
```suggestion
        uint8_t GetOrder(const std::string& targetName, const std::vector<std::vector<std::string>>& orderedTargets) const;
```

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.cpp
```diff
@@ -51,12 +59,39 @@ namespace main_
         builder.WriteUpgradePack(outputFilename, fileSystem);
     }
 
+    uint8_t UpgradePackBuilderFacade::GetOrder(const std::string& targetName, const std::vector<std::vector<std::string>>& orderedTargets)
+    {
+        uint8_t order = 0;
+        for (auto targets : orderedTargets)
```
rjaegers, line 65:
```suggestion
        for (const auto& targets : orderedTargets)
```

upgrade/pack_builder/SupportedTargets.cpp
```diff
@@ -14,38 +14,63 @@ namespace application
...
             targets.mandatory.emplace_back(target);
+            mandatory = false;
+        }
+    }
+
+    void SupportedTargetsBuilder::AddInOrder(const SupportedTargets::Target& target)
+    {
+        if (order)
+        {
+            if (targets.order.size() < *order)
```
oguzcanoguz, line 68: The more proper type here would be a map iso a vector, e.g. std::map<uint8_t, std::vector<Target>> so that you don't have to resize for unused orders. You can just do 
`targets.order[*order].emplace_back(target);`

oguzcanoguz (approved):

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.cpp
```diff
@@ -51,12 +59,45 @@ namespace main_
...
+
+            if (targetPos != targets.second.end())
+                return order;
+        }
+        return infra::none;
+    }
+
+    bool UpgradePackBuilderFacade::CheckIfTargetIsInOrder(const std::string& target, const application::SupportedTargets& supportedTargets)
+    {
+        static uint8_t currentOrderOfTarget = 1;
```
oguzcanoguz, line 78: Now it is possible to include all uint8 values as order, including 0

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.cpp
```diff
@@ -51,12 +59,45 @@ namespace main_
...
+    infra::Optional<uint8_t> UpgradePackBuilderFacade::GetOrder(const std::string& targetName, const std::map<uint8_t, std::vector<std::string>>& orderedTargets) const
+    {
+        infra::Optional<uint8_t> order(infra::inPlace, 0);
+        for (const auto& targets : orderedTargets)
+        {
+            ++(*order);
+            const auto targetPos = std::find(targets.second.begin(), targets.second.end(), targetName);
+
+            if (targetPos != targets.second.end())
+                return order;
```
oguzcanoguz, line 71: We can return targets.first and we dont need order then

upgrade/pack_builder/test/TestSupportedTargets.cpp
```diff
@@ -35,3 +35,38 @@ TEST(SupportedTargetsTest, should_add_mandatory_targets)
...
+TEST(SupportedTargetsTest, should_add_targets_in_order)
+{
+    application::SupportedTargets targets = application::SupportedTargets::Create()
+                                                .Order(2)
+                                                .AddHex("application")
+                                                .Order(1)
+                                                .AddHex("data");
+
+    EXPECT_EQ(2, targets.OrderOfTargets().size());
+    EXPECT_EQ("data", targets.OrderOfTargets()[0][0]);
```
oguzcanoguz, line 59: https://stackoverflow.com/questions/262853/c-map-access-discards-qualifiers-const

oguzcanoguz (approved):

oguzcanoguz (approved):

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.cpp
```diff
@@ -51,12 +59,42 @@ namespace main_
...
+            const auto targetPos = std::find(targets.second.begin(), targets.second.end(), targetName);
+            if (targetPos != targets.second.end())
+                return infra::MakeOptional(targets.first);
+        }
+        return infra::none;
+    }
+
+    bool UpgradePackBuilderFacade::CheckIfTargetIsInOrder(const std::string& target, const application::SupportedTargets& supportedTargets)
+    {
+        static uint8_t currentOrderOfTarget = 0;
```
richardapeters, line 75: Please make this a member variable instead of a static variable

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.cpp
```diff
@@ -51,12 +59,42 @@ namespace main_
...
+        static const auto& orderedTargets = supportedTargets.OrderOfTargets();
```
richardapeters, line 76:
```suggestion
        const auto& orderedTargets = supportedTargets.OrderOfTargets();
```

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.cpp
```diff
@@ -51,12 +59,42 @@ namespace main_
...
+        const auto orderToAdd = GetOrder(target, orderedTargets);
+        if (orderToAdd)
```
richardapeters, line 78:
```suggestion
        if (orderToAdd != infra::none)
```

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.cpp
```diff
@@ -51,12 +59,42 @@ namespace main_
...
+        for (const auto& targets : orderedTargets)
+        {
+            const auto targetPos = std::find(targets.second.begin(), targets.second.end(), targetName);
+            if (targetPos != targets.second.end())
+                return infra::MakeOptional(targets.first);
+        }
+        return infra::none;
+    }
+
+    bool UpgradePackBuilderFacade::CheckIfTargetIsInOrder(const std::string& target, const application::SupportedTargets& supportedTargets)
```
richardapeters, line 73: "Check..." is a verb; it must do something. Instead, the primary purpose of this function is to decide on a predicate. That should be reflected in the name, so the name should read like a predicate instead of a verb.

```suggestion
    bool UpgradePackBuilderFacade::IsTargetInOrder(const std::string& target, const application::SupportedTargets& supportedTargets)
```

richardapeters, line 73: This function does two things: First, it inspects the order and finds out if a target obeys the rules; second, it updates the current order. Please separate those two things. If you separate them, then you can make the function that returns a bool a const function

magi-arun: Done

magi-arun: Yes, Changed.

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.cpp
```diff
@@ -51,12 +59,46 @@ namespace main_
...
+                return infra::MakeOptional(targets.first);
+        }
+        return infra::none;
+    }
+
+    bool UpgradePackBuilderFacade::IsTargetInOrder(const std::string& target, const std::map<uint8_t, std::vector<std::string>>& orderedTargets) const
+    {
+        const auto orderToAdd = GetOrder(target, orderedTargets);
+
+        if ((orderToAdd != infra::none) && (currentOrderOfTarget > *orderToAdd))
```
richardapeters, line 77:
```suggestion
        if (orderToAdd != infra::none && currentOrderOfTarget > *orderToAdd)
```

upgrade/pack_builder_instantiations/UpgradePackBuilderFacade.cpp
```diff
@@ -51,12 +59,46 @@ namespace main_
...
+                return false;
+        return true;
```
richardapeters, line 79:
```suggestion
        return orderToAdd == infra::none || currentOrderOfTarget <= *orderToAdd;
```

richardapeters (approved):

PR #338, opened 2023-06-27
feat: add ServiceProxyResponseQueue

daantimmer:
ServiceProxyResponseQueue is a generic extension to ServiceProxy that adds an easy to reuse scalable ResponseQueue.
ServiceProxy can only queue one request to send. However there are usecases where more than one requests to send can be requested. This class adds that generic functionality.

Usage example:

```cpp
//MyProto.proto:
//service MyProtoService{
//    option (service_id) = 1;
//    rpc Ping(Nothing) returns (Nothing) { option (method_id) = 1; }
//}
//service MyProtoServiceResponse{
//    option (service_id) = 2;
//    rpc Pong(Nothing) returns (Nothing) { option (method_id) = 1; }
//}

//MyProtoEcho.hpp:
class MyProtoEcho : MyProtoService
{
public:
    void Ping() override;

private:
    services::ServiceProxyResponseQueue<MyProtoServiceResponseProxy>::WithStorage<5> response;
};

//MyProtoEcho.cpp:
void MyProtoEcho::Ping()
{
    response.RequestSend([this]{
        response.Pong();
    });
    MethodDone();
}

```


protobuf/echo/Echo.hpp
```diff
@@ -494,6 +517,42 @@ namespace services
...
+        , container{ container }
+    {}
+
+    template<class TServiceProxy>
+    void ServiceProxyResponseQueue<TServiceProxy>::RequestSend(infra::Function<void()> onRequestGranted)
+    {
+        if (container.full())
+        {
+            return;
+        }
```
richardapeters, line 534:
```suggestion
            return;
```

protobuf/echo/Echo.hpp
```diff
@@ -207,6 +208,28 @@ namespace services
         infra::Function<void()> onGranted;
     };
 
+    template<class TServiceProxy>
```
rjaegers, line 211: TServiceProxy -> ServiceProxyType.

protobuf/echo/Echo.hpp
```diff
@@ -207,6 +208,28 @@ namespace services
...
+        explicit ServiceProxyResponseQueue(Container& container, TArgs&&... args);
+
+        void RequestSend(infra::Function<void()> onRequestGranted) override;
+
+    private:
+        void ProcessSendQueue();
+
+        Container& container;
+
+        bool responseInProgress{ false };
```
rjaegers, line 230:
```suggestion
        void ProcessSendQueue();

    private:
        Container& container;
        bool responseInProgress{ false };
```

rjaegers (changes requested):
Please address the comments as mentioned below.

In addition: please add tests for the new class, and add a description to the pull request about the use case and what the class solves.

protobuf/echo/Echo.hpp
daantimmer: yep, sorry 'bout that.

protobuf/echo/Echo.hpp
daantimmer: Are you sure? I thought we were going to adhere to the sonarlint warnings to not have consecutive same-access specifier?

rjaegers: Yes, I am sure. We are going to disable that Sonar warning. Our coding guidelines clearly describe that we use the access specifiers to distinguish different blocks of code and/or methods and variables.

daantimmer: (y)

rjaegers (approved):

PR #350, opened 2023-07-18
ci: add dependency scanner and FetchContent all external dependencies

rjaegers:
Use cmake-dependency-submission to submit CMake FetchContent'ed dependencies to the GitHub Dependency Submission API.


lwip/lwip-2.1.1/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+# cmake-dependency-scan pkg:github/lwip-tcpip/lwip@3b53b6e481beed5787017222463626ae17943beb
```
EkelmansPh, line 1: This added to notify the action there is a dependency that the scan misses?

upgrade/security_key_generator/MaterialGenerator.cpp
```diff
@@ -1,9 +1,9 @@
 #include "upgrade/security_key_generator/MaterialGenerator.hpp"
-#include "crypto/micro-ecc/uECC.h"
 #include "generated/echo/UpgradeKeys.pb.hpp"
 #include "hal/generic/FileSystemGeneric.hpp"
 #include "infra/stream/StdVectorOutputStream.hpp"
 #include "infra/syntax/ProtoFormatter.hpp"
+#include "uECC.h"
```
EkelmansPh, line 6: It would be nicer if the path could be more explicit.

EkelmansPh (approved):

lwip/lwip-2.1.1/CMakeLists.txt
rjaegers: Correct, lwip is an external dependency that we don't consume via FetchContent. This annotation/comment makes sure the Action is still able to pick it up.

upgrade/security_key_generator/MaterialGenerator.cpp
rjaegers: Since we are used to that, true. But for mbedtls, args and all other FetchContent'd dependencies we have also stripped the path to what the package itself provides. So, I prefer to keep it this way. It would also make switching to another package source (vcpkg, Conan) easier, since this is the colloquial path.

PR #390, opened 2023-08-21
chore: solved build error with TracingEcho::PrintField being too greedy with enum types

daantimmer:
Added SFINEA to solve a overload of PrintField beeing too greedy on enum types. Two templated versions exist, one to print any type of containers, and one to print a value directly.

The container overload version was too greedy, trying to print an enum type, thus resulting in a compilation error because you can't iterate over an enum.

### DIFF
--- a/protobuf/echo/TracingEcho.hpp
+++ b/protobuf/echo/TracingEcho.hpp
@@ -49,7 +49,7 @@ namespace services
     {}
 
     template<class T>
-    void PrintField(const T& value, services::Tracer& tracer, typename T::template Type<0>* = 0)
+    void PrintField(const T& value, services::Tracer& tracer, typename std::enable_if<!std::is_enum<T>::value>::type* = 0)
     {
         tracer.Continue() << "{";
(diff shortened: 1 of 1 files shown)


protobuf/echo/TracingEcho.hpp
```diff
@@ -49,15 +49,15 @@ namespace services
     {}
 
     template<class T>
-    void PrintField(const T& value, services::Tracer& tracer, typename T::template Type<0>* = 0)
+    void PrintField(const T& value, services::Tracer& tracer, typename std::enable_if<!std::is_enum<T>::value>::type* = 0)
```
richardapeters, line 52: The "T::template Type<0>* = 0" check checks for the presence of a Type map, which should make this function specific to messages. The new "std::enable_if<!std::is_enum<T>::value>::type* = 0" (which I expect should not be necessary, because enums don't have a Type member) is weaker; it just checks that it is not an enum, not that it is a Message. If this check is really necessary (for faulty compilers), then I'd rather have both checks, and a comment explaining why the second check is necessary.

PR #440, opened 2023-10-20
feat: added SerialCommunicationLoopback

daantimmer:
Added SerialCommunicationLoopback that can be used to link together two classes depending on `hal::SerialCommuncation`. The SendData from server will send the data to the client's ReceiveData callback and vice versa.

The actual sending is scheduled via the EventDispatcher to mimick the asynchronous behaviour of a serial communication port.

### DIFF
--- a/services/util/SerialCommunicationLoopback.hpp
+++ b/services/util/SerialCommunicationLoopback.hpp
@@ -9,7 +9,7 @@ namespace services
         : public hal::SerialCommunication
     {
     public:
-        explicit SerialCommunicationLoopbackPeer(SerialCommunicationLoopbackPeer& other);
+        explicit SerialCommunicationLoopbackPeer(SerialCommunicationLoopbackPeer* other);
 
         void SendData(infra::ConstByteRange data, infra::Function<void()> actionOnCompletion) override;
         void ReceiveData(infra::Function<void(infra::ConstByteRange data)> dataReceived) override;
@@ -25,12 +25,14 @@ namespace services
     class SerialCommunicationLoopback
     {
     public:
+        SerialCommunicationLoopback();
+
         SerialCommunicationLoopbackPeer& Server();
         SerialCommunicationLoopbackPeer& Client();
 
     private:
-        SerialCommunicationLoopbackPeer server{ client };
-        SerialCommunicationLoopbackPeer client{ server };
+        SerialCommunicationLoopbackPeer server;
+        SerialCommunicationLoopbackPeer client;
     };
 }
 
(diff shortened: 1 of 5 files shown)


services/util/test/TestSerialCommunicationLoopback.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "infra/util/ByteRange.hpp"
+#include "infra/util/Function.hpp"
+#include "infra/util/test_helper/MockCallback.hpp"
+#include "services/util/SerialCommunicationLoopback.hpp"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+class SerialCommunicationLoopbackTest
+    : public testing::Test
+    , public infra::ClockFixture
```
richardapeters, line 11: This ClockFixture should not be necessary

services/util/test/TestSerialCommunicationLoopback.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+public:
+    services::SerialCommunicationLoopback SerialCommunicationLoopback;
+    services::SerialCommunicationLoopbackPeer& server{ SerialCommunicationLoopback.Server() };
+    services::SerialCommunicationLoopbackPeer& client{ SerialCommunicationLoopback.Client() };
+};
```
richardapeters, line 17: Can you hook up two SerialCommunicationMocks to the server and client, and test functionality via those mocks?

services/util/test/TestSerialCommunicationLoopback.cpp
richardapeters: Instead, please use EventDispatcherFixture

richardapeters (approved):

services/util/test/TestSerialCommunicationLoopback.cpp
daantimmer: Instead of, or as additional tests?

services/util/SerialCommunicationLoopback.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "hal/interfaces/SerialCommunication.hpp"
+
+namespace services
+{
+    class SerialCommunicationLoopbackPeer
+        : public hal::SerialCommunication
+    {
+    public:
+        explicit SerialCommunicationLoopbackPeer(SerialCommunicationLoopbackPeer* other);
```
richardapeters, line 12:
```suggestion
        explicit SerialCommunicationLoopbackPeer(SerialCommunicationLoopbackPeer& other);
```

PR #441, opened 2023-10-20
fix: added workaround mbedtls and clang-cl incomatibility

daantimmer:
Added workaround by replacing CMAKE_C_COMPILER_ID with CMAKE_C_SIMULATE_ID when configuring mbedtls

### DIFF
--- a/external/crypto/mbedtls/CMakeLists.txt
+++ b/external/crypto/mbedtls/CMakeLists.txt
@@ -11,7 +11,27 @@ set(ENABLE_TESTING Off CACHE INTERNAL "")
 set(ENABLE_PROGRAMS Off CACHE INTERNAL "")
 set(GEN_FILES Off CACHE INTERNAL "")
 
+# mbedtls uses CMAKE_C_COMPILER_ID to determine the type of compiler used.
+# When compiling with xwin, and thus clang-cl, mbedtls thinks that the clang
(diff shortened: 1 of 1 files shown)


external/crypto/mbedtls/CMakeLists.txt
```diff
@@ -11,8 +11,28 @@ set(ENABLE_TESTING Off CACHE INTERNAL "")
...
+endif()
+
 FetchContent_MakeAvailable(mbedtls)
 
+set(CMAKE_C_COMPILER_ID ${ORIGINAL_C_COMPILER_ID})
+set(CMAKE_CXX_COMPILER_ID ${ORIGINAL_CXX_COMPILER_ID})
+
 function(add_mbedtls_target_properties)
     foreach(target ${ARGN})
         target_compile_options(${target} PUBLIC
```
BarisTanyeri, line 38: I think you can avoid setting a global variable. You could add the required variables as compile options for mbedTLS. 

For example it would be: 
-DCMAKE_C_COMPILER_ID=${CMAKE_C_SIMULATE_ID}
-DCMAKE_CXX_COMPILER_ID=${CMAKE_CXX_SIMULATE_ID}

BarisTanyeri (approved):
Aligned offline, agreed on current solution (y)

PR #457, opened 2023-11-10
feat: added generic gatt dis characteristic uuid

daantimmer:
These UUIDs are generic to BLE.
Also fixed the location of services/network/gap.proto, which should have been services/ble/gap.proto instead


EkelmansPh (approved):

services/ble/Gatt.hpp
```diff
@@ -37,6 +37,20 @@ namespace services
         };
     };
 
+    namespace uuid
+    {
+        constexpr static AttAttribute::Uuid16 deviceInformationService{ 0x180A };
```
richardapeters, line 42: Please check if constexpr inline is a better choice

richardapeters (approved):

services/ble/Gatt.hpp
```diff
@@ -37,6 +37,20 @@ namespace services
         };
     };
 
+    namespace uuid
```
rjaegers, line 40: @oguzcanoguz I believe these were previously (re)moved. What was the rationale back then, and is that still relevant now?

oguzcanoguz: Rationale was that we shall not put all those standard services in infralib. DIS is one and there can be surely others later.

magi-arun: DIS is generic service and used widely unlike other standard services which are specific for the application, DIS can be part of ble services in infralib.

daantimmer: @rjaegers @oguzcanoguz @magi-arun let me know if this issue can be resolved as won't fix, or if additional work needs to be done?

magi-arun: As DIS is widely used BLE service, agreed to retain the dis uuid  as part of infralib. Closing this comment as won't fix.

PR #464, opened 2023-11-15
chore: update amp-devcontainer from v2.5.0 to v4.0.2

rjaegers:


.devcontainer/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM ghcr.io/philips-software/amp-devcontainer:v4.0.2@sha256:2d052e0bcb3840ea45aaa21ea6904f6d84b0e310c97e5690a6d48da522ec384e
```
daantimmer, line 1: why is this pinned to version _and_ sha?

rjaegers: Pinning to _both_ version _and_ sha has several benefits:
- It's directly clear what version is used without needing a comment
- It's immutable
- Dependabot understands this format to include release/changelog information in pull requests it generates to update the container

PR #474, opened 2023-11-20
feat: optional tracing

oguzcanoguz:


services/tracer/Tracer.cpp
```diff
@@ -6,19 +6,35 @@ namespace services
...
     {
         return stream;
     }
+#else
+    Tracer::EmptyTracing Tracer::Trace()
+    {
+        return EmptyTracing{};
+    }
+
+    infra::TextOutputStream Tracer::Continue()
```
richardapeters, line 27: Why doesn't Continue return EmptyTracing?

richardapeters (approved):

PR #502, opened 2023-12-13
fix: remove virtual destructors reducing binary size

rjaegers:
Implementing static analysis recommendations of adding virtual to destructors grew binary sizes of upstream projects.


richardapeters:
I think we should disable the SonarCloud warning about virtual functions without a virtual destructor

richardapeters (approved):

PR #503, opened 2023-12-13
fix: remove virtual destructors reducing binary size

oguzcanoguz:


BarisTanyeri (approved):
Looks good. How about the Gatt.hpp and GattServer.hpp?

PR #523, opened 2024-01-11
feat!: refactor ADC interface

gabrielsantosphilips:


hal/interfaces/AnalogToDigitalPin.hpp
```diff
@@ -28,11 +28,21 @@ namespace hal
         virtual void Measure(const infra::Function<void(int32_t value)>& onDone) = 0;
     };
 
+    template<class Storage>
+    class AnalogToDigitalBulkSamples
+    {
+    public:
+        ~AnalogToDigitalBulkSamples() = default;
```
richardapeters, line 35:
```suggestion
        AnalogToDigitalBulkSamples() = default;
        AnalogToDigitalBulkSamples(const AnalogToDigitalBulkSamples& other) = delete;
        AnalogToDigitalBulkSamples& operator=(const AnalogToDigitalBulkSamples& other) = delete;
        ~AnalogToDigitalBulkSamples() = default;
        
```

hal/interfaces/AnalogToDigitalPin.hpp
```diff
@@ -28,11 +28,21 @@ namespace hal
         virtual void Measure(const infra::Function<void(int32_t value)>& onDone) = 0;
     };
 
+    template<class Storage>
+    class AnalogToDigitalBulkSamples
```
richardapeters, line 32: This interface is templated, and therefore flexible towards usage by clients, similar to AnalogToDigitalPin. However, this makes it much more difficult to implement by a chip-specific implementation: how is a concrete implementation going to cater for a wide range of Storage types? You can consider modeling the combination of AnalogToDigitalPin and AnalogToDigitalPinBase more closely, with a AnalogToDigitalPinConverter that converts from platform/chip-specific ADC measured values towards business-logic level values

gabrielsantosphilips: I don't think it makes the concrete implementation that difficult, check how I implemented on amp-hal-st: https://github.com/philips-software/amp-hal-st/pull/238/files#diff-a56557833ff364d67c0f6778e0871831d31a9a8d4197fc73030cee685fded882

Actually I tried to combine in the classes already implemented (https://github.com/philips-software/amp-embedded-infra-lib/pull/523/commits/659458daa70e88265f58b133154246adfa977784). However it is not suitable because of the AnalogToDigitalPinConverter, we have a mismatch in the MemoryRange (one class expects a MemoryRange<Storage> and the another one expects a MemoryRange<infra::Quantity<T, U>>)

gabrielsantosphilips:
@richardapeters  I'm closing this PR because the commits are too messy (merges, rebases and reversions). This is the new one: https://github.com/philips-software/amp-embedded-infra-lib/pull/530

PR #524, opened 2024-01-12
fix: ServiceForwarder with LimitedStreamReader ends up in infinite loop

EkelmansPh:


protobuf/echo/test/TestServiceForwarder.cpp
```diff
@@ -1,8 +1,10 @@
+#include "infra/stream/LimitedInputStream.hpp"
 #include "infra/stream/StdVectorInputStream.hpp"
 #include "infra/stream/StdVectorOutputStream.hpp"
 #include "protobuf/echo/ServiceForwarder.hpp"
 #include "protobuf/echo/test_doubles/EchoMock.hpp"
 
+
```
EkelmansPh, line 7:
```suggestion
```

richardapeters (approved):

PR #549, opened 2024-02-04
chore: add FlashHeterogenous

EkelmansPh:


hal/interfaces/FlashHeterogeneous.cpp
```diff
@@ -0,0 +1,53 @@
...
+        T totalSize = 0;
+        for (T sector = 0; sector != sectors.size(); ++sector)
+        {
+            totalSize += sectors[sector];
+            if (address < totalSize)
+                return sector;
+        }
+
+        assert(address == totalSize);
+        return static_cast<T>(sectors.size() - 1);
```
richardapeters, line 35: If address == totalSize, then the address is one past the last byte of flash, so the sector should be one past the last sector as well
```suggestion
        return static_cast<T>(sectors.size());
```

hal/interfaces/FlashHomogeneous.cpp
```diff
@@ -17,18 +17,22 @@ namespace hal
...
+        assert(sectorIndex < numberOfSectors);
         return sizeOfEachSector;
     }
 
     template<class T>
     T FlashHomogeneousBase<T>::SectorOfAddress(T address) const
     {
-        return address / sizeOfEachSector;
+        auto sectorIndex = address / sizeOfEachSector;
+        assert(sectorIndex < numberOfSectors);
```
richardapeters, line 28: This assertion limits the address to be below the end address, which is not consistent with the SectorOfAddress implementation of FlashHeterogenous

richardapeters (approved):

rjaegers (approved):

hal/interfaces/FlashHomogeneous.cpp
```diff
@@ -17,18 +18,22 @@ namespace hal
...
         return sizeOfEachSector;
     }
 
     template<class T>
     T FlashHomogeneousBase<T>::SectorOfAddress(T address) const
     {
-        return address / sizeOfEachSector;
+        auto sectorIndex = address / sizeOfEachSector;
+        assert(sectorIndex <= numberOfSectors);
+        return std::clamp<T>(sectorIndex, 0, numberOfSectors);
```
richardapeters, line 30: Why use std::clamp here? For debug, there's already an assertion. For release, if someone puts in a wrong address there's no need to make thing look a little less dirty

hal/interfaces/test/TestFlashHeterogeneous.cpp
```diff
@@ -0,0 +1,57 @@
...
+    : public testing::Test
+{
+public:
+    std::array<uint32_t, 3> sectorSizes{ 3, 4, 8 };
+    FlashHeterogeneousImpl flash{ infra::MakeRange(sectorSizes) };
+};
+
+TEST_F(FlashHeterogeneousTest, number_of_sectors)
+{
+    ASSERT_EQ(3, flash.NumberOfSectors());
```
richardapeters, line 30:
```suggestion
    EXPECT_EQ(3, flash.NumberOfSectors());
```
This and all tests below should really be EXPECTs, not ASSERTs

PR #572, opened 2024-02-21
feat: add Segger RTT tracing output option

rjaegers:
Add the option (default off) to trace to Segger RTT.

### DIFF
--- a/external/segger_rtt/CMakeLists.txt
+++ b/external/segger_rtt/CMakeLists.txt
@@ -8,16 +8,16 @@ FetchContent_Declare(
 
 FetchContent_MakeAvailable(segger_rtt)
 
-add_library(segger_rtt.rtt ${EMIL_EXCLUDE_FROM_ALL} STATIC)
+add_library(segger.rtt ${EMIL_EXCLUDE_FROM_ALL} STATIC)
 
-target_include_directories(segger_rtt.rtt PUBLIC
+target_include_directories(segger.rtt PUBLIC
     "$<BUILD_INTERFACE:${segger_rtt_SOURCE_DIR}>"
 )
 
-target_sources(segger_rtt.rtt PRIVATE
+target_sources(segger.rtt PRIVATE
     ${segger_rtt_SOURCE_DIR}/Config/SEGGER_RTT_Conf.h
     ${segger_rtt_SOURCE_DIR}/RTT/SEGGER_RTT.c
     ${segger_rtt_SOURCE_DIR}/RTT/SEGGER_RTT.h
 )
 
-emil_exclude_from_clang_format(segger_rtt.rtt)
+emil_exclude_from_clang_format(segger.rtt)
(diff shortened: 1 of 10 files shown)


external/segger_rtt/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+    segger_rtt
+    GIT_REPOSITORY https://github.com/SEGGERMicro/RTT
+    GIT_TAG ff57c3d516ea59b0a2d4d5cafb33a48572b09c0b # v7.54
+    CONFIGURE_COMMAND ""
+    BUILD_COMMAND ""
+)
+
+FetchContent_MakeAvailable(segger_rtt)
+
+add_library(segger_rtt.rtt ${EMIL_EXCLUDE_FROM_ALL} STATIC)
```
richardapeters, line 11: I'm looking at the double mention of "rtt". Is it really rtt that is being delivered by the segger_rtt package? The source files provide a stream on rtt, and a tracer on that stream. How about segger_rtt.tracer?

external/segger_rtt/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+FetchContent_Declare(
+    segger_rtt
+    GIT_REPOSITORY https://github.com/SEGGERMicro/RTT
+    GIT_TAG ff57c3d516ea59b0a2d4d5cafb33a48572b09c0b # v7.54
+    CONFIGURE_COMMAND ""
```
richardapeters, line 5: Do these empty commands do anything?

services/tracer/StreamWriterOnSeggerRtt.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+    void StreamWriterOnSeggerRtt::Insert(infra::ConstByteRange range, infra::StreamErrorPolicy& errorPolicy)
+    {
+        range.shrink_from_back_to(Available());
+        SEGGER_RTT_Write(bufferIndex, range.cbegin(), range.size());
+    }
+
+    std::size_t StreamWriterOnSeggerRtt::Available() const
+    {
+        return SEGGER_RTT_GetAvailWriteSpace(bufferIndex);
```
richardapeters, line 18: How much is this typically? Can I configure this somewhere?

richardapeters (approved):

external/segger_rtt/CMakeLists.txt
rjaegers: Yes, they make sure CMake doesn't try to configure or build this dependency. Since there is no CMake support in the SEGGERMicro/RTT project.

So we use FetchContent here to clone the repository more or less.

services/tracer/StreamWriterOnSeggerRtt.cpp
rjaegers: Default is 1k, and can be overriden from the SEGGER_RTT_Conf.h by preprocessor defines.

E.g. BUFFER_SIZE_UP

external/segger_rtt/CMakeLists.txt
rjaegers: Yeah, that annoyed me as well. I don't know if tracer covers it though. segger_rtt.communication? The library facilitates both up- and down channels.

rjaegers: Or just segger.rtt?

PR #591, opened 2024-03-19
ci: execute static analysis, fuzzing, and dependency scanner on pull request target

richardapeters:


rjaegers (approved):

rjaegers (changes requested):
I think the workflows changed in this PR don't actually run in the context of this PR anymore.

daantimmer (commented):
Are we sure we want this?

> `pull_request_target` event grants workflows triggered by pull requests from forks access to repository secrets and a read/write GITHUB_TOKEN. That is inherently risky if the workflow inadvertently exposes these secrets or allows for unauthorized modifications to the repository.
> https://stackoverflow.com/a/74959635

rjaegers:
Handling of forks will need a different solution.

PR #598, opened 2024-03-29
feat: HttpClientBasic will now reset the contentError flag before a new connection is attempted

daantimmer:


services/network/test/TestHttpClientBasic.cpp
```diff
@@ -12,29 +11,24 @@ class HttpClientBasicMock
     : public services::HttpClientBasic
 {
 public:
-    using services::HttpClientBasic::HttpClientBasic;
+    virtual ~HttpClientBasicMock() = default;
```
richardapeters, line 14: Why do you declare this destructor?

richardapeters (approved):

PR #602, opened 2024-04-11
feat: implement persisted mbed tls session

BarisTanyeri:
Persisted TLS sessions in configuration store


services/network/MbedTlsSession.hpp
```diff
@@ -0,0 +1,167 @@
...
+        infra::BoundedVector<uint8_t>::WithMaxSize<32> identifier;
+    };
+
+    class TlsSessionHasher
+    {
+    public:
+        TlsSessionHasher() = default;
+        TlsSessionHasher(const TlsSessionHasher& other) = delete;
+        TlsSessionHasher& operator=(const TlsSessionHasher& other) = delete;
+        virtual ~TlsSessionHasher() = default;
```
richardapeters, line 53:
```suggestion
        ~TlsSessionHasher() = default;
```

services/network/Network.proto
```diff
@@ -59,3 +59,12 @@ message WiFiAccessPointWithPassword
     string ssid = 1 [(string_size) = 32];
     string key = 2 [(string_size) = 64];
 }
+
+message MbedTlsPersistedSession
+{
+    fixed32 address = 1;
+    string hostname = 2 [(string_size) = 256];
```
richardapeters, line 66: Is this hostname needed?

services/network/Network.proto
```diff
@@ -59,3 +59,12 @@ message WiFiAccessPointWithPassword
...
+    bool clientSessionObtained = 3;
+    bytes serializedSession = 4 [(bytes_size) = 512];
```
richardapeters, line 68: Does mbedTLS (or any other source) give an upper boundary for a session size? Or what would happen if the session size exceeds the storage available?

BarisTanyeri: I haven't found any reference regarding the serialized size. From the experience with our end points, I've seen that a session without a ticket would be around 150-250 bytes and 350-450 bytes with the ticket. (512 because e2). If it doesn't fit we will really_assert :)

services/network/Network.proto
BarisTanyeri: Nope, I forgot to remove it :( Thanks!

richardapeters (approved):

services/network/Network.proto
BarisTanyeri: It has been agreed offline that we will proceed with the current size definition and make the changes if necessary when the session ticket size is strictly defined. Modifications have been made to skip saving the session resumption ticket in to the flash if the determined size does not fit the allocated flash memory.

PR #605, opened 2024-04-16
chore: services/network_instantiations/NetworkAdapter: Add ExecuteUntil and NetworkActivity

richardapeters:


services/network_instantiations/NameLookup.cpp
```diff
@@ -47,6 +47,12 @@ namespace services
         nameLookup.erase(result);
     }
 
+    bool NameLookup::Active() const
+    {
+        std::lock_guard<std::mutex> lock(mutex);
```
EkelmansPh, line 52: Sonar reports 2 issues about this lock_guard. Please resolve before merging PR

services/network_instantiations/EventDispatcherWithNetworkBsd.cpp
EkelmansPh, line 1: Please update the PR title to follow the conventional commits format.

services/network_instantiations/EventDispatcherWithNetworkBsd.cpp
```diff
@@ -61,6 +61,11 @@ namespace services
         datagrams.push_back(datagram);
     }
 
+    bool EventDispatcherWithNetwork::OpenConnections() const
```
EkelmansPh, line 64: Possibly rename to 'HasOpenConnections'. I saw the name 'OpenConnections' first and was expecting this to return the current number of open connections before seeing the return type and implementation.

EkelmansPh (approved):

services/network_instantiations/EventDispatcherWithNetworkBsd.cpp
richardapeters: I've renamed to ConnectionsOpen(), so that it isn't a verb, but can be used without confusion as "if (ConnectionsOpen())"

PR #640, opened 2024-05-28
chore: enable link group rescan

gabrielsantosphilips:
Feature link group rescan is not enabled for arm none eabi gcc toolchain, so this PR enables it.

Use case:
On amp-hal-st there is a new microcontroller that uses static library and triggers a circular dependency. And that can be fixed by using link groups.


rjaegers:
Have you exhausted all other options to fix this in amp-hal-st. Like using object libraries for some of the violating targets? I have done my utmost best to not need link group rescan when doing the "modern CMake" transition. Imho it would be a shame to enable it again if it can be solved differently instead.

gabrielsantosphilips:
> Have you exhausted all other options to fix this in amp-hal-st. Like using object libraries for some of the violating targets? I have done my utmost best to not need link group rescan when doing the "modern CMake" transition. Imho it would be a shame to enable it again if it can be solved differently instead.

Actually, I believe I can resolve it on hal-st. However, I think the neatest solution would be within embedded-infra, as sometimes we need to manage static libraries with cross dependencies.

I will try to solve it there.

richardapeters (approved):

PR #677, opened 2024-07-12
fix: UartUnix receiveData callback being called after destruction

BarisTanyeri:
Due to threading, object might be in destruction while running is true

### DIFF
--- a/hal/unix/UartUnix.cpp
+++ b/hal/unix/UartUnix.cpp
@@ -73,7 +73,7 @@ namespace hal
                 throw std::system_error(EFAULT, std::system_category());
 
             std::unique_lock lock(mutex);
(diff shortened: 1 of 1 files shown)


hal/unix/UartUnix.cpp
```diff
@@ -73,7 +73,8 @@ namespace hal
                 throw std::system_error(EFAULT, std::system_category());
 
             std::unique_lock lock(mutex);
-            receivedData(infra::ConstByteRange(range.begin(), range.begin() + size));
+            if (receivedData)
```
richardapeters, line 76:
```suggestion
            if (receivedData != nullptr)
```

richardapeters (approved):

PR #688, opened 2024-08-05
feat: gap proto extension before sesame

oguzcanoguz:


services/ble/Gap.proto
```diff
@@ -31,6 +50,26 @@ message SecurityLevel
     SecurityLevelEnum level = 1;
 }
 
+message SecurityModeAndLevel
+{
+    enum SecurityLevelEnum
+    {
+        None = 0;
```
EkelmansPh, line 57: I see different casing styles used for enums. The coding standard mentions enum values should use camelCase.

services/ble/Gatt.proto
```diff
@@ -18,3 +18,112 @@ message DisServiceData
...
+
+message Characteristic
+{
+    bytes uuid = 1 [(bytes_size) = 16];
+    Handle handle = 2;
+    Handle valueHandle = 3;
+    uint32 properties = 4;
+}
+
+message Descriptr
```
EkelmansPh, line 65:
```suggestion
message Descriptor
```

services/ble/Gatt.proto
```diff
@@ -18,3 +18,112 @@ message DisServiceData
...
+    rpc ServiceDiscovered(Service) returns (Nothing) { option (method_id) = 2; }
+    rpc ServiceDiscoveryComplete(Nothing) returns (Nothing) { option (method_id) = 3; }
+    
+    rpc CharacteristicDiscovered(Characteristic) returns (Nothing) { option (method_id) = 4; }
+    rpc CharacteristicDiscoveryComplete(Nothing) returns (Nothing) { option (method_id) = 5; }
+
+    rpc DescriptorDiscovered(Descriptr) returns (Nothing) { option (method_id) = 6; }
+    rpc DescriptorDiscoveryComplete(Nothing) returns (Nothing) { option (method_id) = 7; }
+
+    rpc ReadComplete(ReadResult) returns (Nothing) { option (method_id) = 8; }
```
EkelmansPh, line 123: ReadComplete, but no Read rpc in GattClient?

EkelmansPh (approved):

services/ble/Gatt.proto
DKalkotwar: After discussion, Checked and added.

services/ble/Gap.proto
DKalkotwar: Ignoring for now.

oguzcanoguz: @EkelmansPh 
Correct, it is inconsistent even in this file.
I am not willing to alter the enums in the messages used by old central since I want to avoid updating it. 
But we can be consistent among the rest of the fields.

services/ble/Gatt.proto
oguzcanoguz: See the later commint. Descriptor keyword is not allowed.

PR #699, opened 2024-08-30
feat: Generate string literals

PatrickVKM:
String literals are available at compile time.
This way it is possible for the compiler/linker to place them in a known location for version retrieval of a boot-loader or other applications in the flash memory


PatrickVKM (commented):
Just added some fields.

dcayy (approved):

PR #712, opened 2024-09-18
fix: services.echo_console: parsing tokens of nested messages and arrays, formatting of messages, and properly close upon losing connection

richardapeters:


services/echo_console/Console.cpp
```diff
@@ -1002,112 +1011,115 @@ namespace application
...
+                    throw ConsoleExceptions::IncorrectType{ valueIndex, "string" };
 
                 formatter.PutStringField(infra::BoundedConstString(value.Get<std::string>().data(), value.Get<std::string>().size()), field.number);
             }
 
             void VisitEnum(const EchoFieldEnum& field) override
             {
                 if (!value.Is<int64_t>())
-                    throw ConsoleExceptions::IncorrectType{ valueIndex };
+                    throw ConsoleExceptions::IncorrectType{ valueIndex, "integer" };
```
EkelmansPh, line 1078: Possibly create an explicit ConsoleExceptions::IncorrectIntegerType exception? (and string/bool/etc...)

EkelmansPh (approved):

services/echo_console/Console.cpp
richardapeters: I think there are too many different IncorrectTypes; especially were the type is field.protoType

PR #718, opened 2024-09-28
build: add option EMIL_FETCH_ECHO_COMPILERS to avoid building the compilers even on host

richardapeters:


protobuf/protoc_echo_plugin/test/CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
+if (EMIL_FETCH_ECHO_COMPILERS)
```
BarisTanyeri, line 1: If you were to revert the logic, I think you wouldn't need this workaround

BarisTanyeri (approved):

protobuf/protoc_echo_plugin/test/CMakeLists.txt
richardapeters: Yeah... but then I would adjust the interface to accommodate the implementation, that feels backwards

PR #723, opened 2024-10-03
chore: add ble state 'initiating'

gabrielsantosphilips:
This PR adds the missing GAP state "initiate" as mentioned in the BLUETOOTH CORE SPECIFICATION Version 5.3 (Vol 6, Part B, 1.1 LINK LAYER STATES). This state specifically impacts BLE Central and is unrelated to other roles.

image

The "initiate" state allows for detection of connection failures when no events are returned from the stack, which can block further stack operations.

Two potential scenarios for this state are:

Attempting to connect to a device that does not exist (e.g., due to a wrong MAC address).
The peripheral crashing or entering an undefined state between scanning and connecting, regardless of the cause.

The BLUETOOTH CORE SPECIFICATION Version 5.3 | Vol 6, Part B, 4.5 CONNECTION STATE explains in detail how a connection works:

image


oguzcanoguz:
I am not 100% aligned on the problems PR is trying to solve.

> 1. Attempting to connect to a device that does not exist (e.g., due to a wrong MAC address).
> 2. The peripheral crashing or entering an undefined state between scanning and connecting, regardless of the cause.

In both cases, stack should not be able so send CONNECT_IND to the peer and we should not receive connection complete event. 
So we are in the initiating state and cannot switch to connected state.
What happens when this is the case? Don't we get any event at all?
So these cases does not reflect the figure:
image

The case described in the above figure seems different.
There we are able to go to the connected state but not able to exhange any data packets.
In this case, ble stack should eventually emit a disconnected event.
Is the timing of this event too long? Is it related to supervision timeout?

services/ble/test/TestGapCentral.cpp
```diff
@@ -180,9 +182,10 @@ namespace services
         services::GapState stateScanning = services::GapState::scanning;
         services::GapState stateAdvertising = services::GapState::advertising;
         services::GapState stateConnected = services::GapState::connected;
+        services::GapState stateInitiating = services::GapState::initiating;
 
-        stream << stateStandby << " " << stateScanning << " " << stateAdvertising << " " << stateConnected;
+        stream << stateStandby << " " << stateScanning << " " << stateAdvertising << " " << stateConnected << " " << stateInitiating;
```
daantimmer, line 187: this whole function can be simplified to;

```cpp
    TEST(GapInsertionOperatorStateTest, state_overload_operator)
    {
        infra::StringOutputStream::WithStorage<128> stream;

        stream << services::GapState::standby << " " << services::GapState::scanning << " " << services::GapState::advertising << " " << services::GapState::connected << " " << services::GapState::initiating;

        EXPECT_EQ("Standby Scanning Advertising Connected Initiating", stream.Storage());
    }
```

gabrielsantosphilips:
@oguzcanoguz I updated the picture on https://github.com/philips-software/amp-hal-st/pull/420 to reflect the problem correctly as we just discussed.

oguzcanoguz:
> @oguzcanoguz I updated the picture on [philips-software/amp-hal-st#420](https://github.com/philips-software/amp-hal-st/pull/420) to reflect the problem correctly as we just discussed.

oguzcanoguz:
@gabrielsantosphilips Thanks. 
Now I better understand the issue. 
I also can see why ST ble stack cannot determine a time-out itself: The stack does not know about the advertising interval of the peripheral. It can be from 20ms to 10.24 seconds. And it can also be a noisy environment where not all the advertising packets reach the central. It cannot know when to stop listening. 

Just for this reason, I recommend also not having a fixed time-out inside our GapCentral implementation. 
I'd suggest adding an optional time-out argument to GapCentral::Connect. 
I'd also add a comment for the Connect method to briefly expain that calling connect will first trigger a state update with Initiating and then either another update to Connected or to Standby if there is time-out.

services/ble/Gap.proto
```diff
@@ -107,6 +108,7 @@ message ConnectionParameters
 {
     Address address = 1;
     AddressType addressType = 2;
+    uint32 initiatingTimeout = 3;
```
oguzcanoguz, line 111: The time unit?

services/ble/Gap.hpp
```diff
@@ -328,7 +330,7 @@ namespace services
         void StateChanged(GapState state) override;
 
         // Implementation of GapCentral
-        void Connect(hal::MacAddress macAddress, GapDeviceAddressType addressType) override;
+        void Connect(hal::MacAddress macAddress, GapDeviceAddressType addressType, infra::Duration initiatingTimeout) override;
```
oguzcanoguz, line 333: Will disconnect cancel a initiating state, or does the user have to wait for the timeout?

oguzcanoguz (approved):

services/ble/Gap.hpp
gabrielsantosphilips: No, a disconnect function will fail with 0x3c (command disallowed), but no event will be triggered towards the application. To my opinion a disconnect should **not** cancel a connection, because it should act after the connection is established.

gabrielsantosphilips: I will add a new function to Cancel a connection request.

richardapeters:
BEGIN_COMMIT_OVERRIDE
feat!: add ble state 'initiating'
END_COMMIT_OVERRIDE

PR #766, opened 2024-11-08
feat: add query and fragment parsing to HttpRequestParser

richardapeters:


services/network/test/TestHttpServer.cpp
```diff
@@ -10,6 +10,19 @@
 #include "services/network/test_doubles/HttpServerMock.hpp"
 #include "gmock/gmock.h"
 
+TEST(HttpRequestParserTest, parser_elements)
```
BarisTanyeri, line 13: Can you also add a test without the query parameter? i.e GET a/b#fragment

BarisTanyeri (approved):

PR #780, opened 2024-11-27
feat: replaced assert with really_assert to align `BoundedString` overflow behaviour with other `Bounded` classes

daantimmer:
All other `Bounded` classes really_assert on modifying methods that either under or overflow. `BoundedString` only had a normal assert, which is optimized away from all but Debug builds by default. This could result in actual buffer overflows and underflows destroying neighbouring memory.

This change aligns the `BoundedString` over/underflow behaviour with the other `Bounded` classes.


BarisTanyeri (commented):
Assert may have been used intentionally here. This class (or any other core class) is being used very often and will most probably will result in much larger binaries and few ticks slower. This is probably safer but has significant impact. Approving this PR would mean this change can be applied to other classes, which will have a much larger impact. After this consideration, I'm more leaning towards the optimized out assert because micro tests should in theory catch these failures.

richardapeters (approved):

PR #781, opened 2024-11-29
feat: refactor BodyReader handling in HttpClient implementations

BarisTanyeri:
This also fixes crashes when data is consumed asyncronously and TracingHttpClient* is being used to trace the body of the response

### DIFF
--- a/services/network/HttpClientImpl.cpp
+++ b/services/network/HttpClientImpl.cpp
@@ -222,9 +222,11 @@ namespace services
             else
             {
                 bodyReader.Emplace(ConnectionObserver::Subject().ReceiveStream(), *contentLength);
+                auto reader = infra::MakeContainedSharedObject(bodyReader->countingReader, bodyReaderAccess.MakeShared(bodyReader));
+                BodyReaderAvailable(reader);
 
                 if (HttpClient::IsAttached())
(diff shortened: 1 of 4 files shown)


services/network/HttpClientImpl.cpp
```diff
@@ -222,9 +222,11 @@ namespace services
             else
             {
                 bodyReader.Emplace(ConnectionObserver::Subject().ReceiveStream(), *contentLength);
+                auto reader = infra::MakeContainedSharedObject(bodyReader->countingReader, bodyReaderAccess.MakeShared(bodyReader));
+                BodyReaderAvailable(reader);
 
                 if (HttpClient::IsAttached())
-                    Observer().BodyAvailable(infra::MakeContainedSharedObject(bodyReader->countingReader, bodyReaderAccess.MakeShared(bodyReader)));
+                    Observer().BodyAvailable(std::move(reader));
```
richardapeters, line 229: Please move this into BodyReaderAvailable, and have descendant implementations call their inherited BodyReaderAvailable

richardapeters (approved):

services/network/HttpClientImpl.cpp
BarisTanyeri: Good suggestion! done!

PR #786, opened 2024-12-10
feat: improve hal::Can::Id

fabiangottstein:
This PR makes `hal::Can::Id` creation constexpr and adds defaulted equality operators.
C++ 20 is enabled to make use of default equality operators and is already enabled in `amp-hal-st`.

### DIFF
--- a/hal/interfaces/Can.hpp
+++ b/hal/interfaces/Can.hpp
@@ -13,8 +13,15 @@ namespace hal
         class Id
         {
         public:
-            static Id Create11BitId(uint32_t id);
-            static Id Create29BitId(uint32_t id);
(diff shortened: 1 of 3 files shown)


hal/interfaces/Can.hpp
```diff
@@ -13,17 +13,28 @@ namespace hal
...
+                return Can::Id(id | indicator29Bit);
+            }
 
             bool Is11BitId() const;
             bool Is29BitId() const;
 
             uint32_t Get11BitId() const;
             uint32_t Get29BitId() const;
 
+            bool operator==(const Id& other) const;
```
richardapeters, line 32: Can you add operator!= as well for symmetry?

richardapeters (approved):

PR #802, opened 2025-01-06
fix: assert when serial server and terminal receive data queue is full

magi-arun:


services/network/SerialServer.cpp
```diff
@@ -12,6 +12,7 @@ namespace services
     {
         serialCommunication.ReceiveData([this](infra::ConstByteRange buffer)
             {
+                really_assert(!receiveQueue.Full());
```
richardapeters, line 15: I would rather see a really_assert in QueueForOneReaderOneIrqWriter<T>::AddFromInterrupt, and that it does not just check for Full() but whether buffer.size() fits

richardapeters (approved):

PR #803, opened 2025-01-07
feat: make all hal::Can::Id functions constexpr and add static assert tests

daantimmer:


hal/interfaces/Can.hpp
```diff
@@ -3,42 +3,36 @@
...
+#include <cassert>
 #include <cstdint>
 
 namespace hal
 {
     class Can
     {
     public:
-        class Id
+        class [[nodiscard]] Id
```
fabiangottstein, line 14: What's the benefit of having a `[[nodiscard]]` class here?

daantimmer: It simply means that whatever returns an Id should never be ignored. For example: https://godbolt.org/z/x8nn5KaGn

fabiangottstein: That I understand, but with this specific class I cannot see concrete benefits since it does not for example manage resources. 
On the other hand the code is bloated with the `[[nodiscard]]` tag. This is very unique on this code base

hal/interfaces/Can.hpp
```diff
@@ -3,42 +3,36 @@
...
         {
         public:
-            static constexpr Id Create11BitId(uint32_t id)
-            {
-                return Can::Id(id);
-            }
+            [[nodiscard]] static constexpr Id Create11BitId(uint32_t id);
+            [[nodiscard]] static constexpr Id Create29BitId(uint32_t id);
```
daantimmer, line 18:
```suggestion
            static constexpr Id Create11BitId(uint32_t id);
            static constexpr Id Create29BitId(uint32_t id);
```

fabiangottstein (approved):

Issue #240, opened 2023-03-28
Upgrade Mbed-TLS to version >= 3.3.0 after issue with MSVC builds has been fixed


rjaegers:
Version 3.3.0 of Mbed-TLS does not build with Visual Studio versions < 2022 (i.e. our VS2019 ci-build fails).

This is a known issue in Mbed-TLS registered as: https://github.com/Mbed-TLS/mbedtls/issues/7087.

When this issue is fixed we should upgrade.

ccrugoPhilips:
Brief update on this issue:

Currently, embedded infra lib is pointing to a MbedTls fork containing a provisional patch for the build issue with MSVC v142 (VS 2019) and prior, for mbedtls version 3.6.2.
A more elaborated fix was proposed to the [Mbed-TLS/TF-PSA-Crypto](https://github.com/Mbed-TLS/TF-PSA-Crypto) repository, and is awaiting for review under https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/258. However, if accepted, we would have to wait until next MbedTls major release (4.x.x) to make use of it.
For the above, VS2019 build was removed from the embedded infralib CI, which should allow to point back to MbedTls repository.

Of course, when trying to build with TLS 1.3 configuration enabled, a new platform-specific build (warning as) error appeared (not present with VS 2022 toolset locally, but in WS 2022 only). Though a simple fix for this was added to the PR mentioned above, it brings us back to wait for the next MbedTls release with the patch, if accepted.

For now, will continue pointing to the fork, while keeping an eye on MbedTls and TF-PSA-Crypto repositories for updates.

ccrugoPhilips:
Brief update on this issue:

Currently, embedded infra lib is pointing to a MbedTls fork containing a provisional patch for the build issue with MSVC v142 (VS 2019) and prior, for mbedtls version 3.6.2.
A more elaborated fix was proposed to the [Mbed-TLS/TF-PSA-Crypto](https://github.com/Mbed-TLS/TF-PSA-Crypto) repository, and is awaiting review under https://github.com/Mbed-TLS/TF-PSA-Crypto/pull/258. However, if accepted, we would have to wait until next MbedTls major release (4.x.x) to make use of it.
For the above, VS2019 build was removed from the embedded infralib CI, which should allow to point back to MbedTls repository.

ccrugoPhilips:
[...]

PR #813, opened 2025-01-27
feat: update mbedtls from v3.2.1 to v3.6.2

ccrugoPhilips:
Updating mbedtls for full TLS 1.3 support.


rjaegers:
Please note that there are still several issues with Visual Studio builds with new(er) versions of mbedtls.

See: #240

ccrugoPhilips:
As per @rjaegers. Windows build fails due to [https://github.com/Mbed-TLS/mbedtls/issues/7087](url)

external/crypto/mbedtls/mbedtls_emil_config.h
```diff
@@ -1018,9 +1199,7 @@
  *
  * Uncomment this macro to prevent loading of default entropy functions.
  */
-#ifndef EMIL_HOST_BUILD
```
richardapeters, line 1021: We probably want to keep this define

services/network/MbedTlsSession.hpp
```diff
@@ -27,13 +27,15 @@ namespace services
         virtual void Reinitialize();
         virtual void Obtained();
         virtual bool IsObtained();
+        virtual bool IsLoaded();
```
BarisTanyeri, line 30:
```suggestion
        virtual bool IsDeserialized();
```

services/network/MbedTlsSession.cpp
```diff
@@ -255,6 +259,12 @@ namespace services
                 {
                     SerializeSessionToFlash(session);
                 });
+
+            if (!storage.back().IsLoaded())
```
BarisTanyeri, line 263:
```suggestion
            if (!storage.back().IsDeserialized())
```

external/crypto/mbedtls/mbedtls_emil_config.h
```diff
@@ -131,7 +131,9 @@
  *
  * Comment if your system does not support time functions
  */
+#ifdef EMIL_HOST_BUILD
 #define MBEDTLS_HAVE_TIME
+#endif
```
BarisTanyeri, line 136: Looks like this configuration was always defined by default. What is the missing linkage that led you to add this ifdef ?

services/network/MbedTlsSession.cpp
```diff
@@ -34,7 +33,7 @@ namespace services
         , identifier(reference.identifier)
     {
         mbedtls_ssl_session_init(&session);
-        really_assert(mbedtls_ssl_session_load(&session, reference.serializedSession.begin(), reference.serializedSession.size()) == 0);
+        clientSessionLoaded = mbedtls_ssl_session_load(&session, reference.serializedSession.begin(), reference.serializedSession.size()) ? false : true;
```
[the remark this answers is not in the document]
ccrugoPhilips: addressed

services/network/MbedTlsSession.hpp
```diff
@@ -27,13 +27,15 @@ namespace services
...
         virtual int SetSession(mbedtls_ssl_context* context);
         virtual int GetSession(mbedtls_ssl_context* context);
         virtual const infra::BoundedVector<uint8_t>& Identifier() const;
 
     private:
         mbedtls_ssl_session session;
         bool clientSessionObtained = false;
+        bool clientSessionLoaded = false;
```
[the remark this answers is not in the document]
ccrugoPhilips: addressed

services/network/MbedTlsSession.hpp
ccrugoPhilips: addressed

services/network/MbedTlsSession.cpp
ccrugoPhilips: addressed

external/crypto/mbedtls/mbedtls_emil_config.h
ccrugoPhilips: Addressed with Baris. New mbedtls version does require to provide a time source when no os time source available (building for embedded platform). Conditionally enabling MBEDTLS_PLATFORM_MS_TIME_ALT along with the definition of mbedtls_ms_time worked fine.

services/network/MbedTlsSession.cpp
```diff
@@ -34,7 +33,8 @@ namespace services
         , identifier(reference.identifier)
     {
         mbedtls_ssl_session_init(&session);
-        really_assert(mbedtls_ssl_session_load(&session, reference.serializedSession.begin(), reference.serializedSession.size()) == 0);
+        if (mbedtls_ssl_session_load(&session, reference.serializedSession.begin(), reference.serializedSession.size()) == 0)
+            clientSessionDeserialized = true;
```
richardapeters, line 37:
```suggestion
        clientSessionDeserialized = mbedtls_ssl_session_load(&session, reference.serializedSession.begin(), reference.serializedSession.size()) == 0;
```

services/network/MbedTlsSession.cpp
```diff
@@ -255,6 +260,12 @@ namespace services
                 {
                     SerializeSessionToFlash(session);
                 });
+
+            if (!storage.back().IsDeserialized())
+            {
+                nvm->erase(&persistedSession);
```
richardapeters, line 266: You are iterating over nvm and erasing from it. This will cause the iterator to become invalid. Instead, you need to have a loop that allows updating the iterator while erasing

PR #817, opened 2025-02-06
feat: streamed http put and post without chunked transfer encoding

BarisTanyeri:
This PR enables sending POST and PUT requests using a stream with a known content size. 
Previous behaviour was, that either content needed to be fully available or needed to use chunked encoding. In the cases where chunked encoding is not supported, sending data using streams was not possible. Streaming is now enabled with a known content size for POST and PUT.


services/network/HttpClientImpl.hpp
```diff
@@ -138,6 +141,37 @@ namespace services
             bool done = false;
         };
 
+        class SendingStateForwardLimitedSendStream
```
richardapeters, line 144: Can you consider renaming this class? "LimitedSendStream" points to implementation, not usage. What about SendingStateForwardDefinedSize?

richardapeters (approved):

PR #823, opened 2025-02-13
feat: add AreAnyBitsSet() and AreAllBitsSet()

fabiangottstein:


daantimmer (commented):
Please rename AreBitsSet to AreAnyBitsSet and add AreAllBitsSet which compares against the mask instead of not 0.

Add unit tests

fabiangottstein:
> Please rename AreBitsSet to AreAnyBitsSet and add AreAllBitsSet which compares against the mask instead of not 0.
> 
> Add unit tests

Done

PR #824, opened 2025-02-13
feat: add support to resolve MAC address new approach

cassio-lazaro:


protobuf/echo_attributes/CMakeLists.txt
```diff
@@ -4,13 +4,21 @@ if (EMIL_HOST_BUILD)
...
+        libprotobuf
     )
 
     target_compile_definitions(protobuf.echo_attributes PRIVATE
         LIBPROTOC_EXPORTS
     )
 
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+        # Workaround for warning C4141: 'inline': used more than once
+        # Since we set warning as error and protobuf is adding inline twice, the earning C4141 is raised.
```
richardapeters, line 17:
```suggestion
        # Since we set warning as error and protobuf is adding inline twice, the warning C4141 is raised.
```

richardapeters: Why isn't warning C4141 disabled instead?

PR #839, opened 2025-03-27
fix: keep websocket alive while reader or writer is alive

richardapeters:


daantimmer (approved):
General comment on the unit tests:
If you have an EXPECT_CALL like:
```cpp
EXPECT_CALL(instance, Call());
//or
EXPECT_CALL(instance, Call(testing::_));
```
Then you can actually omit the parens/arguments:
```cpp
EXPECT_CALL(instance, Call);
```

PR #850, opened 2025-04-15
feat: updated Gap to allow connection param setting and interval change for ble internal flash

dcayy:
- added SetConnectionParameter method to Gap to allow connection parameter setting from application level
- added GapPeripheralIntervalController interface to allow connection and adv interval change control

### DIFF
--- a/services/ble/Gap.hpp
+++ b/services/ble/Gap.hpp
@@ -223,6 +223,7 @@ namespace services
         using infra::Observer<GapPeripheralObserver, GapPeripheral>::Observer;
 
         virtual void StateChanged(GapState state) = 0;
+        virtual void ConnectionIntervalUpdated(uint16_t connMultiplierMin, uint16_t connMultiplierMax) = 0;
     };
 
     class GapPeripheral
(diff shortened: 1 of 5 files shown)


services/ble/Gap.hpp
```diff
@@ -276,6 +284,7 @@ namespace services
         infra::ConstByteRange GetScanResponseData() const override;
         void Advertise(GapAdvertisementType type, AdvertisementIntervalMultiplier multiplier) override;
         void Standby() override;
+        void SetConnectionParameter(const services::GapConnectionParameters& connParam) override;
```
oguzcanoguz, line 287:
```suggestion
        void SetConnectionParameters(const services::GapConnectionParameters& connParam) override;
```

services/ble/Gap.hpp
```diff
@@ -214,6 +214,13 @@ namespace services
         bool IsDeviceBonded(hal::MacAddress address, GapDeviceAddressType addressType) const override;
     };
 
+    class GapPeripheralIntervalController
```
oguzcanoguz, line 217: I think this interface is rather business specific and not suitable for infralib

oguzcanoguz (approved):

services/ble/Gap.hpp
dcayy: actually yes, i will move it upper

PR #863, opened 2025-05-08
fix: when UartUnix is destroyed receivedData is still called on read thread

BarisTanyeri:
- Established partial locks in Read()
- Reset receivedData on destruction

### DIFF
--- a/hal/unix/UartUnix.cpp
+++ b/hal/unix/UartUnix.cpp
@@ -28,11 +28,11 @@ namespace hal
 
     UartUnix::~UartUnix()
     {
-        LockReadThen([this]
-            {
-                running = false;
-                receivedData = nullptr;
-            });
+        {
+            std::unique_lock lock(receivedDataMutex);
+            running = false;
+            receivedData = nullptr;
+        }
         if (readThread.joinable())
             readThread.join();
     }
@@ -54,10 +54,10 @@ namespace hal
     {
         running = true;
 
-        LockReadThen([this, &dataReceived]
-            {
-                receivedData = dataReceived;
-            });
+        {
+            std::unique_lock lock(receivedDataMutex);
+            receivedData = dataReceived;
+        }
 
         if (!readThread.joinable())
             readThread = std::thread([this]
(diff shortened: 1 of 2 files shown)


daantimmer (dismissed):
Please move the LockReadThen out of the header in to an anonymous namespace in the source file:
```cpp
namespace
{
  void LockReadThen(mutex-ref receivedDataMutex, function action){ 
    std::unique_lock lock(receivedDataMutex);
    action();
  }
}
```

hal/unix/UartUnix.cpp
```diff
@@ -28,7 +28,11 @@ namespace hal
 
     UartUnix::~UartUnix()
     {
-        running = false;
+        LockReadThen([this]
+            {
+                running = false;
+                receivedData = nullptr;
+            });
```
richardapeters, line 35: Please just use the standard idiom

```suggestion
        {
            std::unique_lock lock(receivedDataMutex);
            running = false;
            receivedData = nullptr;
        }
```

hal/unix/UartUnix.cpp
```diff
@@ -50,10 +54,10 @@ namespace hal
...
         running = true;
 
-        {
-            std::unique_lock lock(receivedDataMutex);
-            receivedData = dataReceived;
-        }
+        LockReadThen([this, &dataReceived]
+            {
+                receivedData = dataReceived;
+            });
```
richardapeters, line 60:
```suggestion
        {
            std::unique_lock lock(receivedDataMutex);
            receivedData = dataReceived;
        }
```

hal/unix/UartUnix.cpp
```diff
@@ -66,17 +70,28 @@ namespace hal
...
+                        receivedData(infra::ConstByteRange(range.begin(), range.end()));
+                });
         }
     }
+
+    void UartUnix::LockReadThen(const infra::Function<void()>& action)
+    {
+        std::unique_lock lock(receivedDataMutex);
+        action();
+    }
```
richardapeters, line 96:
```suggestion
```

richardapeters (changes requested):
Having the lock in a separate function was confusing for me. Locking a lock and first statement in a block is the expected thing to do

BarisTanyeri:
So essentially, for this specific occasion Sonar was incorrect. Without the function maintainability was B. I refactored it in to a function, maintainability is now A with the function, but in reality function makes it more confusing...

BarisTanyeri:
> Please move the LockReadThen out of the header in to an anonymous namespace in the source file:
> 
> ```c++
> namespace
> {
>   void LockReadThen(mutex-ref receivedDataMutex, function action){ 
>     std::unique_lock lock(receivedDataMutex);
>     action();
>   }
> }
> ```

I'd rather have it in the class if I was going to go for a function, but I think it's more readable without a function...

richardapeters (approved):

PR #891, opened 2025-06-12
feat: add MakeConstRange which returns a MemoryRange<const T>

richardapeters:


oguzcanoguz (approved):

infra/util/MemoryRange.hpp
```diff
@@ -231,7 +237,7 @@ namespace infra
     template<class U>
     bool MemoryRange<T>::operator==(const MemoryRange<U>& rhs) const
     {
-        return begin() == rhs.begin() && end() == rhs.end();
+        return (empty() && rhs.empty()) || (begin() == rhs.begin() && end() == rhs.end());
```
danielschenk, line 240: Is this actually faster? Cause empty needs to check both begins and ends too, eventually. Anyway, guess the compiler has more options to choose the quickest path now. Might be interesting to look at this with Compiler Explorer?

richardapeters: It's not about speed; two empty ranges with different begin and end pointers should still compare equal

PR #897, opened 2025-06-19
feat: automatically enable coverage for all targets excluding test targets

daantimmer:


.github/workflows/static-analysis.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           GTEST_OUTPUT: "xml:${{ github.workspace }}/testresults/"
       - name: Collect coverage
         run: |
-          gcovr --sonarqube=coverage.xml --exclude-lines-by-pattern '.*assert\(.*\);|.*really_assert\(.*\);|.*std::abort();' --exclude-unreachable-branches --exclude-throw-branches -j "$(nproc)" --exclude=.*/generated/.* --exclude=.*/examples/.* --exclude=.*/external/.* --exclude=.*/lwip/.* --exclude=.*/tracing/.* --exclude=.*/test/.*
+          gcovr --sonarqube=coverage.xml --exclude-lines-by-pattern '.*assert\(.*\);|.*really_assert\(.*\);|.*std::abort();' --exclude-unreachable-branches --exclude-throw-branches -j "$(nproc)" --exclude=.*/generated/.* --exclude=.*/examples/.* --exclude=.*/external/.* --exclude=.*/lwip/.* --exclude=.*/tracing/.*
```
BarisTanyeri, line 42: I still think that */test/.* coverage information is not required. Why did you end up removing it ?

daantimmer: Because there is no more coverage data generated for anything inside a /test/ folder

PR #901, opened 2025-06-26
refactor: enable default constructor for BoundedDequeIterator

fabiangottstein:
std::deque defines its iterator as a form of `LegacyForwardIterator` which needs to be default constructable.
In the effort of using `BoundedDeque` with the same semantics as std::deque it is required to enable the default constructor of its iterator.

### DIFF
--- a/infra/util/BoundedDeque.hpp
+++ b/infra/util/BoundedDeque.hpp
@@ -156,7 +156,7 @@ namespace infra
             using pointer = value_type*;
             using reference = value_type&;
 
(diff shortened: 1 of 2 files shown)


fabiangottstein:
Windows build issues originate from a compiler migration from `MSVC 19.43.34808.0` to `MSVC 19.44.35209.0`.
The github runners ship this compiler with the `Visual Studio Enterprise 2022 | 17.14.36203.30` package.

PR #904, opened 2025-07-07
fix(deps): some warnings when building for Windows on Linux

danielschenk:
- protobuf: don't pass unsupported /MP flag to MSVC frontend for clang
- protobuf: fix incorrectly disabled invalid-offsetof warning on clang
- include all external library headers as system headers, so we don't have to disable warnings for our own compilation units


external/protobuf/CMakeLists.txt
```diff
@@ -48,5 +48,19 @@ if (EMIL_HOST_BUILD)
...
+                set_target_properties(${target} PROPERTIES COMPILE_OPTIONS "${all_options}")
+            endif()
+        endforeach()
+    endfunction()
+
     add_protobuf_target_properties(libprotobuf libprotoc)
+
+    # /MP is added by protobuf in case of MSVC, but that check is a (partial) false positive since we use the MSVC _frontend_ for clang
+    # which does not support this flag
+    remove_mp(libprotobuf libprotobuf-lite libprotoc libupb)
```
richardapeters, line 65: But there is also a configuration with plain old MSVC, so /MP should not be removed unconditionally

external/protobuf/CMakeLists.txt
```diff
@@ -48,5 +48,19 @@ if (EMIL_HOST_BUILD)
         endforeach()
     endfunction()
 
+    function(remove_mp)
```
richardapeters, line 51: What about merging remove_mp into add_protobuf_target_properties, and renaming that to modify_protobuf_target_properties?

PR #906, opened 2025-07-09
fix: fix CucumberStepStorage (variable type)

gabrielsantosphilips:


magi-arun:
PR name can be more descriptive about changes and avoid double fix

magi-arun (approved):

PR #907, opened 2025-07-09
feat: rename Pair to PairAndBond

XavierBernhardt-Philips:
Renamed pair in gap.proto to PairAndBond

### DIFF
--- a/services/ble/Gap.hpp
+++ b/services/ble/Gap.hpp
@@ -3,7 +3,6 @@
 
 #include "hal/interfaces/MacAddress.hpp"
 #include "infra/timer/Timer.hpp"
-#include "infra/util/BoundedString.hpp"
 #include "infra/util/BoundedVector.hpp"
 #include "infra/util/ByteRange.hpp"
 #include "infra/util/EnumCast.hpp"


services/ble/Gap.hpp
```diff
@@ -251,6 +255,10 @@ namespace services
         virtual void Advertise(GapAdvertisementType type, AdvertisementIntervalMultiplier multiplier) = 0;
         virtual void Standby() = 0;
         virtual void SetConnectionParameters(const services::GapConnectionParameters& connParam) = 0;
+
+    protected:
+        virtual void StartedAdvertising(infra::BoundedConstString functionName) {};
```
oguzcanoguz: These protected methods should not be part of the abstract interface. If they are for logging, which I think is the case, please have them in the implementation class which the Tracing class is extending.

XavierBernhardt-Philips: Fixed.

services/ble/Gap.hpp
```diff
@@ -3,6 +3,7 @@
 
 #include "hal/interfaces/MacAddress.hpp"
 #include "infra/timer/Timer.hpp"
+#include "infra/util/BoundedString.hpp"
```
richardapeters:
```suggestion
```

richardapeters (approved):

PR #911, opened 2025-07-12
fix: fix cucumber wire protocol controller class

gabrielsantosphilips:


BarisTanyeri (approved):
LGTM! I wonder if this would have been caught by valgrind

PR #912, opened 2025-07-16
[DRAFT] feat: adapt sc revert legacy bleui definition mapping

MiquelJayson-Philips:
This is part of the following PR: https://github.com/philips-internal/amp-connectivity-nodes/pull/1896 .


services/ble/Gap.hpp
```diff
@@ -128,13 +128,24 @@ namespace services
             level4,
         };
 
-        virtual void Pair() = 0;
+        // Values align with STM32 sdk ble core defs
```
gabrielsantosphilips: I don't think we should align with a specific manufacturer. if for some reason the manufacturer has a different values, the class implement this interface should be responsible for adapting it.

services/ble/Gap.hpp
```diff
@@ -128,13 +128,24 @@ namespace services
...
+        enum class SecureConnectionMode : uint8_t
```
oguzcanoguz, line 131: I think we have the same three options for mitm and secureconnections:
            notSupported,
            supported,
            enforced, 
            
For some vendors, such as ST, we might not have the implementation for all combinations, which is fine.

oguzcanoguz (approved):

MiquelJayson-Philips:
With the recent changes to Gap.proto this is not necessary anymore.

PR #917, opened 2025-07-24
fix: handle connection failure for empty hostname and port in HttpClientCachedConnection for cached connections

BarisTanyeri:
- When a cached connection is invalidated hostname becomes empty, cached connections with empty hostnames should not be delegated further but should be immediately dropped to service clients that do have a valid hostname. 
- If they are not dropped, connect will be delegated to connect to an empty host which will eventually result in a name lookup failure. 
- Without needing to connect we already know that those clients will result in name lookup resolution failure so those clients can be dropped before we open a socket and connect.


richardapeters:
The HttpClientConnector interface (implicitly) promises that Hostname() is valid throughout a Connect() call. If the hostname becomes invalid, it should invoke CancelConnect(). Therefore, the issue seen should be resolved at the calling site, not at HttpClientCachedConnectionConnector

PR #928, opened 2025-08-11
fix: service forwarder transfer

magi-arun:


protobuf/echo/ServiceForwarder.cpp
```diff
@@ -69,11 +69,12 @@ namespace services
...
         while (processedSize != forwardingSize && contentsWriter != nullptr && contentsReader != nullptr)
         {
-            auto range = contentsReader->ExtractContiguousRange(forwardingSize - processedSize);
+            auto range = contentsReader->ExtractContiguousRange(std::min<std::size_t>(forwardingSize - processedSize, contentsWriter->Available()));
+
             contentsWriter->Insert(range, errorPolicy);
             processedSize += range.size();
 
-            if (contentsReader->Empty())
+            if (contentsWriter->Empty() || contentsReader->Empty())
```
gabrielsantosphilips, line 77: Can you create a new unit test to verify this behavior?

magi-arun: Will be address here https://tfsemea1.ta.philips.com/tfs/TPC_Region14/Connectivity/_workitems/edit/162391

PR #930, opened 2025-08-15
implement retry mechanism

cassio-lazaro:


EkelmansPh (approved):

oguzcanoguz:
Hi, I see that made an error and left one point unclear in the design doc. 
So I have some corrections. 

1. I just realized that, for a gatt client, BLE_STATUS_INSUFFICIENT_RESOURCES is only returned for "write without response" calls. (See [here](https://github.com/philips-software/amp-hal-st/blob/9591db7500281909fdc48def98f2c21a0558e95c/hal_st/middlewares/STM32_WPAN/STM32CubeWBA/ble/stack/include/auto/ble_events.h#L2361)) So, we only need to have "infra::Function<void(OperationStatus)>& onDone" argument in WriteWithoutResponse method. And only retry WriteWithoutResponse operations in RetryGattClientCharacteristicsOperations class. 
2. For Write method, it is fine to have "infra::Function<void(uint8_t)>& onDone" argument as it can never be retried and might return GATT error code after completion which is still uint8_t. Similar with Read method.
3. WriteWithoutResponse calls do not compete with each other and with other Gatt operations since they are fire-and-forget operations. In contrast, for operations that have a callback from peer such as Read and Write operations, there can be only one active operation. So we don't need to use claimer for WriteWithoutResponse in ClaimingGattClientAdapter.

oguzcanoguz (approved):

PR #931, opened 2025-08-18
fix: sesame secure: first hash data before signing

richardapeters:


richardapeters:
There's already a unit test, and the sign and verify functions test each other. If they're not updated in sync, then the unit test fails

PR #937, opened 2025-08-19
fix: session resumption with tls 1.3

ccrugoPhilips:


external/crypto/mbedtls/mbedtls_emil_config.h
```diff
@@ -1795,7 +1795,7 @@
  *
  * Comment this macro to disable support for TLS 1.2 / DTLS 1.2
  */
-#define MBEDTLS_SSL_PROTO_TLS1_2
```
richardapeters: Is this absolutely necessary? Because this will break TLS connectivity with a lot of peers

richardapeters (approved):

PR #950, opened 2025-09-04
fix: claiming gatt client WriteWithoutResponse crash during Read

danielschenk:
Fixes an issue with `ClaimingGattClientAdapter::WriteWithoutResponse` which caused state of other operations to be reused and overwritten. This could occur when receiving notifications from the peripheral during a pending read, for example, because the notification is acknowledged using a `WriteWithoutResponse`.

fix: unintentional release of claim in `ClaimingGattClientAdapter::WriteWithoutResponse` (#950)
feat: refactor gatt client interface to take handle directly (#950)

### DIFF
--- a/services/ble/ClaimingGattClientAdapter.cpp
+++ b/services/ble/ClaimingGattClientAdapter.cpp
@@ -127,11 +127,7 @@ namespace services
 
     void ClaimingGattClientAdapter::WriteWithoutResponse(AttAttribute::Handle handle, infra::ConstByteRange data, const infra::Function<void(OperationStatus)>& onDone)
     {
-        onWriteWithoutResponseDone = onDone;
-        GattClientObserver::Subject().WriteWithoutResponse(handle, data, [this](OperationStatus result)
-            {
-                onWriteWithoutResponseDone(result);
-            });
+        GattClientObserver::Subject().WriteWithoutResponse(handle, data, onDone);
     }
 
     void ClaimingGattClientAdapter::EnableNotification(AttAttribute::Handle handle, const infra::Function<void(uint8_t)>& onDone)
--- a/services/ble/ClaimingGattClientAdapter.hpp
+++ b/services/ble/ClaimingGattClientAdapter.hpp
@@ -117,7 +117,6 @@ namespace services
 
         std::optional<std::variant<DiscoveredService, DiscoveredCharacteristic, DiscoveredDescriptor, HandleRange>> discoveryContext;
         std::optional<CharacteristicOperation> characteristicOperationContext;
-        infra::Function<void(OperationStatus)> onWriteWithoutResponseDone;
 
         infra::ClaimableResource resource;
         infra::ClaimableResource::Claimer characteristicOperationsClaimer{ resource };


services/ble/GattClient.hpp
```diff
@@ -89,15 +78,14 @@ namespace services
...
-        // Implementation of GattClientCharacteristicOperationsObserver
-        AttAttribute::Handle CharacteristicValueHandle() const override;
+        AttAttribute::Handle CharacteristicValueHandle() const;
 
         // Implementation of GattClientStackUpdateObserver
         void NotificationReceived(AttAttribute::Handle handle, infra::ConstByteRange data) override;
         void IndicationReceived(AttAttribute::Handle handle, infra::ConstByteRange data, const infra::Function<void()>& onDone) override;
 
     private:
+        GattClientCharacteristicOperations* operations{ nullptr };
```
cassio-lazaro, line 88: Why is it a pointer and not a reference? It's being set on the constructor and not changed anymore.

services/ble/GattClient.cpp
```diff
@@ -1,11 +1,12 @@
 #include "services/ble/GattClient.hpp"
+#include "infra/util/ReallyAssert.hpp"
 
 namespace services
 {
     GattClientCharacteristic::GattClientCharacteristic(GattClientCharacteristicOperations& operations, AttAttribute::Uuid type, AttAttribute::Handle handle, AttAttribute::Handle valueHandle, GattCharacteristic::PropertyFlags properties)
         : GattClientCharacteristic(type, handle, valueHandle, properties)
     {
-        GattClientCharacteristicOperationsObserver::Attach(operations);
+        this->operations = &operations;
```
cassio-lazaro, line 9: It's better to initialize in the _initialization list_.

services/ble/GattClient.hpp
danielschenk: It's set only by one of the two available constructors. You don't need to create this with the operations available necessarily

services/ble/GattClient.cpp
danielschenk: Agree, but the compiler didn't allow me. I think it's because the other constructor (without operations) is called in the initializer list

services/ble/ClaimingGattClientAdapter.cpp
```diff
@@ -125,61 +125,58 @@ namespace services
             });
     }
 
-    void ClaimingGattClientAdapter::WriteWithoutResponse(const GattClientCharacteristicOperationsObserver& characteristic, infra::ConstByteRange data, const infra::Function<void(OperationStatus)>& onDone)
+    void ClaimingGattClientAdapter::WriteWithoutResponse(AttAttribute::Handle handle, infra::ConstByteRange data, const infra::Function<void(OperationStatus)>& onDone)
     {
-        characteristicOperationContext.emplace(WriteWithoutResponseOperation{ data, onDone }, characteristic.CharacteristicValueHandle());
-        const auto& writeWithoutResponseContext = std::get<WriteWithoutResponseOperation>(characteristicOperationContext->operation);
-        GattClientObserver::Subject().WriteWithoutResponse(*this, writeWithoutResponseContext.data, [this](OperationStatus result)
+        onWriteWithoutResponseDone = onDone;
```
oguzcanoguz: Do we need to store this or can we just pass along?

services/ble/GattClient.cpp
danielschenk: For more info see: https://stackoverflow.com/a/12190126/8271554

The workaround in question here is not possible as our full constructor takes a reference and not a pointer. Also I don't like nullptr being in the initializer-list, it should be default next to the member declaration.

PR #958, opened 2025-09-15
feat: extend ble gatt server interface to add characteristic descriptors

brianvanderbeek-philips:
Relates to: https://github.com/philips-software/amp-hal-st/pull/625

### DIFF
--- a/services/ble/GattServer.hpp
+++ b/services/ble/GattServer.hpp
@@ -60,7 +60,23 @@ namespace services
         // on unrecoverable failure (i.e. BLE stack indicates an issue
         // with updating or sending data).
         virtual UpdateStatus Update(const GattServerCharacteristicOperationsObserver& characteristic, infra::ConstByteRange data) const = 0;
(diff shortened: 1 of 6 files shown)


services/ble/GattServer.hpp
```diff
@@ -89,6 +90,8 @@ namespace services
         uint16_t ValueLength() const;
         uint8_t GetAttributeCount() const;
 
+        virtual void AddDescriptor(const AttAttribute::Uuid& uuid, infra::ConstByteRange data) = 0;
```
oguzcanoguz: Lets please make it explicit here that this is a static descriptor.
AddStaticDescriptor?

brianvanderbeek-philips: Is that not confusing, given that in BLE terms there is no 'static characteristic descriptor' ? Also in terms of variable names there is no distinction between const, static, non-static, so I would expect function names also not to make this distinction.

oguzcanoguz: Is ReadOnly better?

brianvanderbeek-philips: outdated. Code has been changed.

PR #974, opened 2025-10-01
chore: ignore uninitialized-const-pointer in mbedtls

andjordan:


external/crypto/mbedtls/CMakeLists.txt
```diff
@@ -43,10 +43,13 @@ function(add_mbedtls_target_properties)
             # see https://github.com/Mbed-TLS/mbedtls/pull/6966
             # mbedtls sets the -Wdocumentation flag, which is throwing warnings
             # since clang-15
-            # Warning treated as error when implicit conversion within mbedtls
             $<$<CXX_COMPILER_ID:Clang>:-Wno-documentation -Wno-conversion>
             $<$<CXX_COMPILER_ID:AppleClang>:-Wno-documentation -Wno-conversion>
+            # Warning treated as error when implicit conversion within mbedtls
             $<$<CXX_COMPILER_ID:GNU>:-Wno-conversion>
+            # Warnings treated as error since clang-22
```
richardapeters, line 50: I don't think we need the conditional on the clang version. Just ignoring the warning is enough

richardapeters (approved):

external/crypto/mbedtls/CMakeLists.txt
andjordan: It is required because the mutation build is on clang-18, which doesn't have these new warning flags

PR #1060, opened 2026-01-07
feat!: remove function device started from ble proto files

gabrielsantosphilips:


services/ble/Gap.proto
```diff
@@ -274,12 +274,11 @@ service GapPeripheralResponse
     option (service_id) = 34;
 
     rpc Done(Nothing) returns (Nothing) { option (method_id) = 1; }
-    rpc DeviceStarted(Nothing) returns (Nothing) { option (method_id) = 2; }
```
heinwessels-philips, line 277: I still need to have a discussion with Richard on when our proto files should maintain clean backwards compatibility. But until then I believe it would be better to do it just in case. So could you maintain the old `method_id`s, and reserved the removed one? For example:

```suggestion
    reserved 2; 
    reserved "DeviceStarted";
    rpc StateChanged(State) returns (Nothing) { option (method_id) = 3; }
```

services/ble/Gap.proto
```diff
@@ -292,9 +291,8 @@ service GapCentralResponse
     rpc PairingSuccessfullyCompleted(Nothing) returns (Nothing) { option (method_id) =  4; }
     rpc PairingFailed(PairingStatus) returns (Nothing) { option (method_id) = 5; }
     rpc NumberOfBondsChanged(UInt32Value) returns (Nothing) { option (method_id) = 6; }
-    rpc DeviceStarted(Nothing) returns (Nothing) { option (method_id) = 7; }
```
heinwessels-philips, line 295: Ditto

heinwessels-philips (changes requested):

services/ble/Gap.proto
gabrielsantosphilips: It doesn't make sense to me. It isn't compatible anyway—it uses a different proto file and a different service.

gabrielsantosphilips: Oh, sorry—I thought your comment was on a different repo. Here, I do agree with your comment.

PR #1151, opened 2026-03-26
feat: add SingleInstance and NotDestructable interfaces

heinwessels-philips:
This adds two interfaces: `infra::SingleIntance` and `infra::NotDestructable` that other classes can derive from to inheret those properties.

We have many instances where we write classes that we assume will never be destructed or only ever have one instance, but this is never enforced. And if those assumptions ever break it can cause hard to track bugs. For example:
- use-after-free crashes using event dispatcher in a destructed object, which causes a crash recently at a client. 
- Another case where I wanted it recently is in: #1148

So now you can write
```C++
class MySuperImportantClass
     : infra::NotDestructable
     , infra::SingleInstance<MySuperImportantClass>
{}
```

And if those assumptions are ever broken then it will immediately crash with  a clear error message. The hope is that notifies the developer during development something was used outside it's intended bounds. And not waiting until possible crashes occur later as a side effect of that.

## Why not just delete destructor?
About `NotDestructable`, ideally this could be done at compile time by simply removing the destructor. These classes are usually defined `static` anyway right, so it should compile? But I could not find a way to get this to work, not even to speak about portability. So this runtime check is deemed good enough for now.


infra/util/SingleInstance.hpp
```diff
@@ -0,0 +1,41 @@
...
+namespace infra
+{
+    template<class Tag>
+    class SingleInstance
+    {
+        static int& NumberOfInstances()
+        {
+            static int numberOfInstances = 0;
+            return numberOfInstances;
+        }
```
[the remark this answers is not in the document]
heinwessels-philips: Thread safety is not currently a requirement

infra/util/SingleInstance.hpp
```diff
@@ -0,0 +1,46 @@
...
```
daantimmer: This could just be a static member variable

infra/util/NotDestructible.hpp
```diff
@@ -0,0 +1,27 @@
+#ifndef INFRA_UTIL_NOT_DESTRUCTIBLE_HPP
+#define INFRA_UTIL_NOT_DESTRUCTIBLE_HPP
+
+#include "infra/util/ReallyAssert.hpp"
+
+namespace infra
+{
+    class NotDestructible
```
daantimmer, line 8: Rename to something not 'NotDestructible', like: `DestructionGuard`

infra/util/NotDestructible.hpp
```diff
@@ -0,0 +1,27 @@
...
+#ifdef EMIL_HOST_BUILD
+        void AllowDestructionOfNotDestructible()
+        {
+            canSafelyDestruct = true;
+        }
+#endif
+    };
+}
+
+#endif // INFRA_UTIL_NOT_DESTRUCTIBLE_HPP
```
daantimmer, line 27: removing closing comment

infra/util/NotDestructible.hpp
```diff
@@ -0,0 +1,27 @@
+#ifndef INFRA_UTIL_NOT_DESTRUCTIBLE_HPP
+#define INFRA_UTIL_NOT_DESTRUCTIBLE_HPP
+
+#include "infra/util/ReallyAssert.hpp"
+
+namespace infra
+{
+    class NotDestructible
+    {
+        bool canSafelyDestruct = false;
```
daantimmer, line 10: Move to a private block below the public functions

infra/util/SingleInstance.hpp
```diff
@@ -0,0 +1,43 @@
...
+#define INFRA_UTIL_SINGLE_INSTANCE_HPP
+
+#include "infra/util/ReallyAssert.hpp"
+
+namespace infra
+{
+    template<class Tag>
+    class SingleInstance
+    {
+        static int numberOfInstances;
```
daantimmer, line 11: Move to a private block below the public functions

daantimmer, line 11: ensure numberOfInstances is thread safe.

daantimmer, line 11: Possibly make numberOfInstances a type of bool, `isActive` for instance. No need to count higher than 1 anyway because program will be terminated

infra/util/SingleInstance.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+    public:
+        SingleInstance()
+        {
+            numberOfInstances++;
+            really_assert_with_msg(numberOfInstances <= 1, "Only single instance allowed");
```
daantimmer, line 17: The robot is giving me this as an example (when changed to std::atomic<bool>):
```cpp
const bool expected = false;
if (!active.compare_exchange_strong(expected, true, std::memory_order_acquire, std::memory_order_relaxed))
    really_assert_with_msg(false, "Only single instance allowed");
```

and the destructor:

```cpp
active.store(false, std::memory_order_release);
```

infra/util/SingleInstance.hpp
```diff
@@ -0,0 +1,43 @@
...
+            numberOfInstances = 0;
+        }
+#endif
+    };
+
+    template<class Tag>
+    int SingleInstance<Tag>::numberOfInstances = 0;
+}
+
+#endif // INFRA_UTIL_SINGLE_INSTANCE_HPP
```
daantimmer, line 43: removing closing comment

infra/util/NotDestructible.hpp
daantimmer, line 8: Do note, a 'NotDestructible' would still make sense, but then in the form of a wrapper, just like google test's NiceMock/NaggyMock/StrictMock like so:

```cpp
NotDestructable<MyImplementation> impl;
```

PR #1164, opened 2026-04-13
feat: version exchange in sesame init/init response

richardapeters:
See synopsis

### DIFF
--- a/services/tracer/TracingEchoInstantiationSecured.cpp
+++ b/services/tracer/TracingEchoInstantiationSecured.cpp
@@ -33,7 +33,7 @@ namespace main_
     {}
 
     TracingEchoOnSesameSecuredDiffieHellman::TracingEchoOnSesameSecuredDiffieHellman(infra::BoundedVector<uint8_t>& cobsSendStorage, infra::BoundedDeque<uint8_t>& cobsReceivedMessage,
-        infra::BoundedDeque<uint8_t>& windowedReceivedMessage, infra::BoundedVector<uint8_t>& securedSendBuffer, infra::BoundedVector<uint8_t>& securedReceiveBuffer, hal::BufferedSerialCommunication& serialCommunication, services::MethodSerializerFactory& serializerFactory, const services::EchoPolicyDiffieHellman::Crypto& crypto, infra::ConstByteRange dsaCertificate, infra::ConstByteRange rootCaCertificate, hal::SynchronousRandomDataGenerator& randomDataGenerator, const services::EchoErrorPolicy& echoErrorPolicy, services::Tracer& tracer, services::SesameInitializer & initializer)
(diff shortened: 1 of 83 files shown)


services/util/SesameWindowed.cpp
```diff
@@ -295,14 +297,14 @@ namespace services
 
     void SesameWindowed::StateSendingInitResponse::Request()
     {
-        communication.SesameEncodedObserver::Subject().RequestSendMessage(3);
+        communication.SesameEncodedObserver::Subject().RequestSendMessage(3 + communication.sesameInitializer.InitInformation().size());
```
heinwessels-philips, line 300: Could this magic number be put into a `static constexpr foo` with an explanatory name? And possibly have a helper functio that can return `foo + init.size()` with an explanatory name? Seems this `3 + foo` is used a lot.

services/util/test/TestSesameWindowed.cpp
```diff
@@ -528,7 +542,6 @@ TEST_F(SesameWindowedTest, Reset_forwards_to_cobs_and_requests_initialize)
 TEST_F(SesameWindowedTest, hold_initialization_until_initializer_grants)
```
heinwessels-philips, line 554: Do we maybe need a test where `initInfo` is empty?

services/tracer/TracingEchoInstantiationSecured.cpp
```diff
@@ -3,8 +3,8 @@
 
 namespace main_
 {
-    TracingEchoOnSesameSecured::TracingEchoOnSesameSecured(infra::BoundedVector<uint8_t>& cobsSendStorage, infra::BoundedDeque<uint8_t>& cobsReceivedMessage, infra::BoundedDeque<uint8_t>& windowedReceivedMessage, infra::BoundedVector<uint8_t>& securedSendBuffer, infra::BoundedVector<uint8_t>& securedReceiveBuffer,
-        hal::BufferedSerialCommunication& serialCommunication, services::MethodSerializerFactory& serializerFactory, const services::SesameSecured::KeyMaterial& keyMaterial, const services::EchoErrorPolicy& echoErrorPolicy, services::Tracer& tracer, services::SesameInitializer& initializer)
+    TracingEchoOnSesameSecured::TracingEchoOnSesameSecured(infra::BoundedVector<uint8_t>& cobsSendStorage, infra::BoundedDeque<uint8_t>& cobsReceivedMessage,
```
heinwessels-philips, line 6: These constructor arguments are getting very unreadable. Would be good to change these at some point to instead receive a struct of arguments or something. But probably out of scope for now

heinwessels-philips (commented):
Not a thorough review, just some thoughts. Would be nice though if there was some context in the PR description to better understand what the code is supposed to do :)

services/util/test/TestSesameWindowed.cpp
```diff
@@ -47,19 +47,34 @@ class SesameWindowedTest
...
     void ReceiveInitRequest(uint16_t availableWindow)
     {
         EXPECT_CALL(observer, Initialized());
-        ReceivePacket(infra::ConstructBin().Value<uint8_t>(1).Value<infra::LittleEndian<uint16_t>>(availableWindow).Vector());
+        EXPECT_CALL(initializer, InitInformationReceived(testing::_)).WillOnce(testing::Invoke([this](infra::StreamReaderWithRewinding& initInfoReader)
+            {
+                infra::DataInputStream::WithErrorPolicy stream(initInfoReader);
+                std::vector<uint8_t> data(stream.Available());
+                stream >> infra::MakeRange(data);
+                EXPECT_TRUE(infra::ContentsEqual(infra::MakeRange(initInfo), infra::MakeRange(data)));
```
andjordan:
```suggestion
                EXPECT_THAT(data. testing::ElementsAreArray(initInfo));
```
not tested :)

andjordan (approved):

PR #1167, opened 2026-04-14
feat: harden Echo RequestSend against multiple calls before granted

heinwessels-philips:
Echo can crash in a hard-to-debug way if two clients using the same proxy communicating at the same time, which then overwrites the `onGranted` handler. This results on the `AutoResetFunction onGranterd` to be empty when it's called the second time, causing a crash.

This change adds an `really_assert` so that it instead crashes immediatelly if an `RequestSend` is called while another grant is in progress, which means the offending code will be easily visible in the stack trace.


andjordan (dismissed):

services/echo_console/Console.cpp
```diff
@@ -492,7 +493,7 @@ namespace application
 
     void Console::CancelRequestSend(services::ServiceProxy& serviceProxy)
     {
-        std::abort();
+        LOG_AND_ABORT_NOT_IMPLEMENTED();
     }
```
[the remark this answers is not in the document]
heinwessels-philips: This was already an abort before, this is not a functional change.

protobuf/echo/Echo.cpp
```diff
@@ -18,9 +19,18 @@ namespace services
     {}
 
     ServiceProxy::~ServiceProxy()
+    {
+        CancelRequestSend();
+    }
+
+    void ServiceProxy::CancelRequestSend()
```
richardapeters: With the name "CancelRequestSend", I wouldn't expect the if-statement here. The name indicates that there is a request that must be cancelled

protobuf/echo/Echo.hpp
```diff
@@ -33,11 +33,12 @@ namespace services
     {
     public:
         ServiceProxy(Echo& echo, uint32_t maxMessageSize);
-        ~ServiceProxy();
+        virtual ~ServiceProxy();
```
richardapeters: I think you made this virtual to solve some warning? I would make it protected instead

protobuf/echo/Echo.hpp
```diff
@@ -33,11 +33,12 @@ namespace services
...
 
         Echo& Rpc();
         virtual void RequestSend(infra::Function<void()> onGranted);
         virtual void RequestSend(infra::Function<void()> onGranted, uint32_t requestedSize);
+        void CancelRequestSend();
```
richardapeters: I don't like mixing virtuals and non virtuals :)

richardapeters (approved):

andjordan (approved):

protobuf/echo/Echo.hpp
heinwessels-philips: Yeah, based on this comment by CoPilot. Moved to protected.

> ServiceProxy has virtual methods (e.g., RequestSend/GrantSend) but its destructor is non-virtual (Echo.hpp). Now that the destructor performs logic (CancelRequestSend), deleting a derived proxy via a ServiceProxy* would be undefined behavior and could skip derived cleanup. Consider making ~ServiceProxy virtual (or otherwise preventing polymorphic deletion, e.g., protected non-virtual destructor).

protobuf/echo/Echo.cpp
```diff
@@ -46,6 +48,14 @@ namespace services
         return std::move(methodSerializer);
     }
 
+    void ServiceProxy::CancelRequestSend()
+    {
+        assert(onGranted);
+        onGranted = nullptr;
+        currentRequestedSize = 0;
+        echo.CancelRequestSend(*this);
```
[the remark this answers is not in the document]
heinwessels-philips: Removed assert

protobuf/echo/EchoOnStreams.cpp
```diff
@@ -27,8 +28,26 @@ namespace services
...
+    }
+
+    void EchoOnStreams::Reset()
     {
         ReleaseReader();
+
+        while (!sendRequesters.empty())
+            sendRequesters.front().CancelRequestSend();
+
+        if (sendingProxy != nullptr)
```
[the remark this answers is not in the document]
heinwessels-philips: The assert is removed

PR #1174, opened 2026-04-20
refactor: data loss in QueueForOneReaderOneIrqWriter

ccrugoPhilips:
Fixes and enhancements to the queue implementation. 
- Renamed methods to better reflect the intent
- Removed multiple non-atomic loads from AddFromInterruptUnchecked
- Improved boundaries check. If the queue does not have enough space available to hold the data passed, a `really_assert ` check fails. 
- Removed unnecessary `assert` check

### DIFF
--- a/infra/event/QueueForOneReaderOneIrqWriter.hpp
+++ b/infra/event/QueueForOneReaderOneIrqWriter.hpp
@@ -36,8 +36,8 @@ namespace infra
         T Get();
         infra::MemoryRange<const T> ContiguousRange(uint32_t offset = 0) const;
         void Consume(uint32_t amount);
-        std::size_t Size() const;
-        std::size_t EmptySize() const;
+        std::size_t QueuedCount() const;
+        std::size_t Capacity() const;
 
         T operator[](size_t position) const;
 
@@ -75,7 +75,7 @@ namespace infra
 
     private:
         QueueForOneReaderOneIrqWriter<T>& queue;
-        std::size_t available{ queue.Size() };
+        std::size_t available{ queue.QueuedCount() };
         std::size_t offset{ 0 };
     };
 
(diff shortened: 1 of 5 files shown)


infra/event/QueueForOneReaderOneIrqWriter.hpp
```diff
@@ -36,8 +36,8 @@ namespace infra
         T Get();
         infra::MemoryRange<const T> ContiguousRange(uint32_t offset = 0) const;
         void Consume(uint32_t amount);
-        std::size_t Size() const;
-        std::size_t EmptySize() const;
+        std::size_t QueuedCount() const;
+        std::size_t Capacity() const;
```
andjordan, line 40: I would keep original names, as they are public and might be used on other repos

infra/event/QueueForOneReaderOneIrqWriter.hpp
```diff
@@ -112,31 +112,30 @@ namespace infra
...
         else
-            contentsEnd += copySize;
+            end += copySize;
 
         data.pop_front(copySize);
 
-        assert(begin <= contentsEnd.load() || begin > contentsEnd.load() + data.size());
-        std::copy(data.begin(), data.end(), contentsEnd.load());
-        contentsEnd += data.size();
+        std::copy(data.begin(), data.end(), end);
```
heinwessels-philips, line 139: Is is still possible to write out-of-bounds by accident with a large `data` if the caller calls `AddFromInterruptUnchecked` directly?

infra/event/test/TestQueueForOneReaderOneIrqWriter.cpp
```diff
@@ -149,28 +149,57 @@ TEST_F(QueueForOneReaderOneIrqWriterTest, Size)
...
 }
 
 TEST_F(QueueForOneReaderOneIrqWriterTest, EmptySize)
 {
     queue.emplace(buffer, [this]() {});
-    EXPECT_EQ(sizeof(buffer) - 1, queue->EmptySize());
+    EXPECT_EQ(sizeof(buffer) - 1, queue->Capacity());
+}
+
+TEST_F(QueueForOneReaderOneIrqWriterTest, add_range_asserts_when_insufficient_space)
```
heinwessels-philips, line 176: We need another test, something like
```suggestion
TEST_F(QueueForOneReaderOneIrqWriterTest, add_range_unchecked_does_not_write_out_bounds)
```

heinwessels-philips (changes requested):

infra/event/QueueForOneReaderOneIrqWriter.hpp
ccrugoPhilips: Yes it is still possible, adding a new assert to fail if end atomic leaves buffer boundaries

heinwessels-philips: I don't think this should be a `really_assert`, because that would not be correct if the caller used `AddFromInterruptUnchecked`. Instead I think it should be an `assert` and early return`. So something like (in pseudo code):

```C++
assert(wontWriteOutOfBounds);
if (willWriteOutOfBounds)
{
    writeWhatWillFit();
    return;
}
```
If the user calls `Unchecked` they wouldn't expect a crash on a check I think

infra/event/QueueForOneReaderOneIrqWriter.hpp
richardapeters: I don't mind breaking this interface, but size() (or Size() in this case) is the term for allocated number of elements in a standard container

richardapeters (approved):

infra/event/test/TestQueueForOneReaderOneIrqWriter.cpp
```diff
@@ -167,10 +167,54 @@ TEST_F(QueueForOneReaderOneIrqWriterTest, Size)
...
+    queue->AddFromInterrupt(wrapData);
+
+    EXPECT_EQ(4, queue->Size());
+    EXPECT_EQ(3, queue->Get());
+    EXPECT_EQ(4, queue->Get());
+    EXPECT_EQ(5, queue->Get());
+    EXPECT_EQ(6, queue->Get());
+}
+
+TEST_F(QueueForOneReaderOneIrqWriterTest, add_range_unchecked_does_not_write_out_bounds)
```
heinwessels-philips: So this is a circular buffer right? So there's technically two cases that can overflow:
- `head` is ahead of `tail` in memory
- `tail` is ahead of `head` in memory

And I would expect the `add` function to comply to two things:
- Never write out of bounds
- The head should never overwrite the tail

Do you think it makes sense to add a test for all these configurations? I don't think we're covering all of them. For both the checking function, and the unchecked function? IMO it would be beneficial because there's so many potential edge cases, and many of these have a unique code that only triggers in that condition.

heinwessels-philips (approved):

infra/event/QueueForOneReaderOneIrqWriter.hpp
```diff
@@ -112,31 +112,34 @@ namespace infra
     template<class T>
     void QueueForOneReaderOneIrqWriter<T>::AddFromInterrupt(infra::MemoryRange<const T> data)
     {
-        really_assert(!Full());
+        really_assert(Capacity() - Size() >= data.size());
```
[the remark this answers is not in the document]
ccrugoPhilips: There is already an existing  non-fatal API that addresses exactly this:
AddFromInterruptUnchecked

heinwessels-philips:
This PR could probably also have the title `fix: ...`

ccrugoPhilips:
> This PR could probably also have the title `fix: ...`

Did discuss the pr name with Richard, and agreed that `refactor` better describes the intent of this PR

PR #1175, opened 2026-04-21
feat: add MacResolver interface and lwIP implementation

uxitraHannes:
Introduces a `MacResolver` interface and two LwIP-backed implementations for resolving IP addresses to MAC addresses: one for IPv4 via ARP, one for IPv6 via Neighbor Discovery (ND6).

### Changes

**`services/network/MacResolver.hpp`** (new)
- Abstract `MacResolver` interface with `void Resolve(const IPAddress&, callback)`
- Destructor moved to `protected` and non-virtual; uses `IPAddress` from `Address.hpp` instead of a local alias

**`lwip/lwip_cpp/MacResolverLwIp.hpp/.cpp`** (new)
- `MacResolverRetryHelper` — shared retry logic backed by `infra::RetryFixedInterval`; rejects concurrent calls gracefully instead of asserting
- `ArpMacResolverLwIp` — resolves IPv4 addresses using `etharp_find_addr` / `etharp_request`; immediately returns `nullopt` for non-IPv4 input
- `Nd6MacResolverLwIp` — resolves IPv6 addresses using `nd6_get_next_hop_addr_or_queue`; checks on-link prefix before queuing; immediately returns `nullopt` for non-IPv6 input
- All variant accesses use `std::get_if` with null-checks; no bare `std::get` that would terminate with exceptions disabled
- Aborts if a resolve is already in progress

### DIFF
--- a/lwip/lwip_cpp/MacResolverLwIp.cpp
+++ b/lwip/lwip_cpp/MacResolverLwIp.cpp
@@ -26,6 +26,8 @@ namespace services
         // lwIP has no public API to trigger a standalone ND6 solicitation.
         // We therefore create a zero-length PBUF_REF probe and call
         // nd6_get_next_hop_addr_or_queue(), which drives neighbor discovery.
+        // lwIP determines whether this is a direct on-link neighbor or a
+        // routed next-hop lookup.
         // If queued, lwIP clones this volatile probe, so the caller still
         // releases the original pbuf after the lookup call returns.
         std::optional<hal::MacAddress> Nd6Lookup(const ip6_addr_t& target)
@@ -50,6 +52,8 @@ namespace services
         }
     }
 
+    // ARP is link-local. This resolver requests the MAC for the provided
+    // IPv4 address as-is and does not perform IPv4 route/next-hop lookup.
     void ArpMacResolverLwIp::SendRequest(IPv4Address address)
     {
         ip4_addr_t target;
--- a/lwip/lwip_cpp/MacResolverLwIp.cpp
+++ b/lwip/lwip_cpp/MacResolverLwIp.cpp
@@ -23,6 +23,10 @@ namespace services
             return target;
         }
 
+        // lwIP has no public API to trigger a standalone ND6 solicitation.
+        // We therefore create a probe pbuf and call
+        // nd6_get_next_hop_addr_or_queue(), which queues this probe and
+        // drives neighbor discovery until the cache can be resolved.
         std::optional<hal::MacAddress> Nd6Lookup(const ip6_addr_t& target)
         {
             const u8_t* hwaddrp = nullptr;
--- a/lwip/lwip_cpp/MacResolverLwIp.cpp
+++ b/lwip/lwip_cpp/MacResolverLwIp.cpp
@@ -19,7 +19,7 @@ namespace services
                 PP_HTONL(address[3] | (static_cast<uint32_t>(address[2]) << 16)),
                 PP_HTONL(address[5] | (static_cast<uint32_t>(address[4]) << 16)),
                 PP_HTONL(address[7] | (static_cast<uint32_t>(address[6]) << 16)));
-            ip6_addr_assign_zone(&target, IP6_UNICAST, netif_default);
+            ip6_addr_set_zone(&target, netif_default->ip6_addr->u_addr.ip6.zone);
             return target;
         }
 
--- a/lwip/lwip_cpp/MacResolverLwIp.cpp
+++ b/lwip/lwip_cpp/MacResolverLwIp.cpp
@@ -24,9 +24,10 @@ namespace services
         }
 
         // lwIP has no public API to trigger a standalone ND6 solicitation.
-        // We therefore create a probe pbuf and call
-        // nd6_get_next_hop_addr_or_queue(), which queues this probe and
-        // drives neighbor discovery until the cache can be resolved.
+        // We therefore create a zero-length PBUF_REF probe and call
+        // nd6_get_next_hop_addr_or_queue(), which drives neighbor discovery.
+        // If queued, lwIP clones this volatile probe, so the caller still
+        // releases the original pbuf after the lookup call returns.
         std::optional<hal::MacAddress> Nd6Lookup(const ip6_addr_t& target)
         {
             const u8_t* hwaddrp = nullptr;
@@ -72,13 +73,18 @@ namespace services
         return std::nullopt;
     }
 
-    void ArpMacResolverLwIp::Resolve(IPv4Address address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone)
+    bool ArpMacResolverLwIp::Resolve(IPv4Address address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone)
     {
+        if (retryTimer_.Armed())
+        {
+            return false;
+        }
+
         auto mac = Lookup(address);
         if (mac)
         {
             onDone(mac);
-            return;
+            return true;
         }
 
         pendingAddress_ = address;
@@ -91,6 +97,8 @@ namespace services
             {
                 OnRetry();
             });
+
+        return true;
     }
 
     void ArpMacResolverLwIp::OnRetry()
@@ -115,14 +123,19 @@ namespace services
         }
     }
 
-    void Nd6MacResolverLwIp::Resolve(const IPv6Address& address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone)
+    bool Nd6MacResolverLwIp::Resolve(const IPv6Address& address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone)
     {
+        if (retryTimer_.Armed())
+        {
+            return false;
+        }
+
         auto target = ToLwipIp6(address);
         auto mac = Nd6Lookup(target);
         if (mac)
         {
             onDone(mac);
-            return;
+            return true;
         }
 
         pendingAddress_ = address;
@@ -133,6 +146,8 @@ namespace services
             {
                 OnRetry();
             });
+
+        return true;
     }
 
     void Nd6MacResolverLwIp::OnRetry()
--- a/lwip/lwip_cpp/MacResolverLwIp.cpp
+++ b/lwip/lwip_cpp/MacResolverLwIp.cpp
@@ -99,13 +99,14 @@ namespace services
             return;
         }
 
-        if (--retriesLeft_ == 0)
+        if (retriesLeft_ == 0)
         {
             retryTimer_.Cancel();
             onDone_(std::nullopt);
         }
         else
         {
+            --retriesLeft_;
             SendRequest(pendingAddress_);
         }
     }
@@ -141,10 +142,14 @@ namespace services
             return;
         }
 
-        if (--retriesLeft_ == 0)
+        if (retriesLeft_ == 0)
         {
             retryTimer_.Cancel();
             onDone_(std::nullopt);
+        }
+        else
+        {
+            --retriesLeft_;
         }
     }
 }
--- a/lwip/lwip_cpp/MacResolverLwIp.cpp
+++ b/lwip/lwip_cpp/MacResolverLwIp.cpp
@@ -1,0 +1,150 @@
+#include "lwip/lwip_cpp/MacResolverLwIp.hpp"
+#include "lwip/ip4_addr.h"
+#include "lwip/ip6_addr.h"
+#include "lwip/nd6.h"
+#include "lwip/netif.h"
+#include "lwip/pbuf.h"
+#include "netif/etharp.h"
+#include <algorithm>
+
+namespace services
+{
+    namespace
+    {
+        ip6_addr_t ToLwipIp6(const services::IPv6Address& address)
+        {
+            ip6_addr_t target;
+            IP6_ADDR(&target,
+                PP_HTONL(address[1] | (static_cast<uint32_t>(address[0]) << 16)),
+                PP_HTONL(address[3] | (static_cast<uint32_t>(address[2]) << 16)),
+                PP_HTONL(address[5] | (static_cast<uint32_t>(address[4]) << 16)),
+                PP_HTONL(address[7] | (static_cast<uint32_t>(address[6]) << 16)));
+            ip6_addr_assign_zone(&target, IP6_UNICAST, netif_default);
+            return target;
+        }
+
+        std::optional<hal::MacAddress> Nd6Lookup(const ip6_addr_t& target)
+        {
+            const u8_t* hwaddrp = nullptr;
+            auto* probe = pbuf_alloc_reference(nullptr, 0, PBUF_REF);
+            if (probe == nullptr)
+            {
+                return std::nullopt;
+            }
+
+            auto result = nd6_get_next_hop_addr_or_queue(netif_default, probe, &target, &hwaddrp);
+            pbuf_free(probe);
+
+            if (result >= 0 && hwaddrp != nullptr)
+            {
+                hal::MacAddress mac;
+                std::copy(hwaddrp, hwaddrp + mac.size(), mac.begin());
+                return mac;
+            }
+            return std::nullopt;
+        }
+    }
+
+    void ArpMacResolverLwIp::SendRequest(IPv4Address address)
+    {
+        ip4_addr_t target;
+        IP4_ADDR(&target, address[0], address[1], address[2], address[3]);
+        etharp_request(netif_default, &target);
+    }
+
+    std::optional<hal::MacAddress> ArpMacResolverLwIp::Lookup(IPv4Address address)
+    {
+        ip4_addr_t target;
+        IP4_ADDR(&target, address[0], address[1], address[2], address[3]);
+
+        struct eth_addr* ethaddr = nullptr;
+        const ip4_addr_t* ipaddr = nullptr;
+        if (etharp_find_addr(netif_default, &target, &ethaddr, &ipaddr) >= 0 && ethaddr != nullptr)
+        {
+            hal::MacAddress mac;
+            std::copy(ethaddr->addr, ethaddr->addr + mac.size(), mac.begin());
+            return mac;
+        }
+        return std::nullopt;
+    }
+
+    void ArpMacResolverLwIp::Resolve(IPv4Address address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone)
+    {
+        auto mac = Lookup(address);
+        if (mac)
+        {
+            onDone(mac);
+            return;
+        }
+
+        pendingAddress_ = address;
+        onDone_ = onDone;
+        retriesLeft_ = retries;
+
+        SendRequest(address);
+
+        retryTimer_.Start(retryInterval, [this]
+            {
+                OnRetry();
+            });
+    }
+
+    void ArpMacResolverLwIp::OnRetry()
+    {
+        auto mac = Lookup(pendingAddress_);
+        if (mac)
+        {
+            retryTimer_.Cancel();
+            onDone_(mac);
+            return;
+        }
+
+        if (--retriesLeft_ == 0)
+        {
+            retryTimer_.Cancel();
+            onDone_(std::nullopt);
+        }
+        else
+        {
+            SendRequest(pendingAddress_);
+        }
+    }
+
+    void Nd6MacResolverLwIp::Resolve(const IPv6Address& address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone)
+    {
+        auto target = ToLwipIp6(address);
+        auto mac = Nd6Lookup(target);
+        if (mac)
+        {
+            onDone(mac);
+            return;
+        }
+
+        pendingAddress_ = address;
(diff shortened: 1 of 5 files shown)


lwip/lwip_cpp/MacResolverLwIp.cpp
```diff
@@ -0,0 +1,159 @@
...
+            const u8_t* hwaddrp = nullptr;
+            auto* probe = pbuf_alloc_reference(nullptr, 0, PBUF_REF);
+            if (probe == nullptr)
+            {
+                return std::nullopt;
+            }
+
+            auto result = nd6_get_next_hop_addr_or_queue(netif_default, probe, &target, &hwaddrp);
+            pbuf_free(probe);
+
```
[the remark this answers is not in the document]
uxitraHannes: No, lwip clones the probe when needed, so it always has to be freed to avoid a leak. I added this behavior in the comment.

lwip/lwip_cpp/MacResolverLwIp.hpp
```diff
@@ -0,0 +1,42 @@
...
+        std::optional<hal::MacAddress> Lookup(IPv4Address address) override;
+        void Resolve(IPv4Address address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone) override;
+
+    private:
+        void OnRetry();
+
+        IPv4Address pendingAddress_{};
+        infra::Function<void(std::optional<hal::MacAddress>)> onDone_;
+        infra::TimerRepeating retryTimer_;
+        uint8_t retriesLeft_ = 0;
```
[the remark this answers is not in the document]
uxitraHannes: Added reject (return false) when already resolving.

services/network/MacResolver.hpp
```diff
@@ -0,0 +1,39 @@
...
+    {
+    protected:
+        Nd6MacResolver() = default;
+        Nd6MacResolver(const Nd6MacResolver& other) = delete;
+        Nd6MacResolver& operator=(const Nd6MacResolver& other) = delete;
+        ~Nd6MacResolver() = default;
+
+    public:
+        virtual void Resolve(const IPv6Address& address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone) = 0;
+    };
```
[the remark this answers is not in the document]
uxitraHannes: Added reject (return false) when already resolving.

lwip/lwip_cpp/MacResolverLwIp.cpp
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    std::optional<hal::MacAddress> ArpMacResolverLwIp::Lookup(IPv4Address address)
+    {
+        ip4_addr_t target;
+        IP4_ADDR(&target, address[0], address[1], address[2], address[3]);
+
+        struct eth_addr* ethaddr = nullptr;
+        const ip4_addr_t* ipaddr = nullptr;
+        if (etharp_find_addr(netif_default, &target, &ethaddr, &ipaddr) >= 0 && ethaddr != nullptr)
```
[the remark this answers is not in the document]
uxitraHannes: Okay, added comments to describe the "next-hop" "on-line" semantics.

lwip/lwip_cpp/MacResolverLwIp.cpp
```diff
@@ -0,0 +1,178 @@
...
+                OnRetry();
+            });
+
+        return true;
+    }
+
+    void Nd6MacResolverLwIp::OnRetry()
+    {
+        auto mac = Nd6Lookup(ToLwipIp6(pendingAddress_));
+
```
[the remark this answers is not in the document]
uxitraHannes: Not possible.

services/network/MacResolver.hpp
```diff
@@ -0,0 +1,39 @@
...
+    class ArpMacResolver
+    {
+    protected:
+        ArpMacResolver() = default;
+        ArpMacResolver(const ArpMacResolver& other) = delete;
+        ArpMacResolver& operator=(const ArpMacResolver& other) = delete;
+        ~ArpMacResolver() = default;
+
+    public:
+        virtual void SendRequest(IPv4Address address) = 0;
```
richardapeters: I understand how "Resolve" can function: That's an asynchronous function, so it takes an address, and eventually invokes onDone. But what are the purposes of SendRequest and Lookup?

services/network/MacResolver.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "hal/interfaces/MacAddress.hpp"
+#include "infra/timer/Timer.hpp"
+#include "infra/util/Function.hpp"
+#include "services/network/Address.hpp"
+#include <optional>
+
+namespace services
+{
+    class ArpMacResolver
```
richardapeters: The implementation uses ARP for resolution, but should that be exposed by the name of the interface? Wouldn't "MacResolver" be a better name?

services/network/MacResolver.hpp
```diff
@@ -0,0 +1,39 @@
...
+    protected:
+        ArpMacResolver() = default;
+        ArpMacResolver(const ArpMacResolver& other) = delete;
+        ArpMacResolver& operator=(const ArpMacResolver& other) = delete;
+        ~ArpMacResolver() = default;
+
+    public:
+        virtual void SendRequest(IPv4Address address) = 0;
+        virtual std::optional<hal::MacAddress> Lookup(IPv4Address address) = 0;
+        virtual bool Resolve(IPv4Address address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone) = 0;
```
richardapeters: Would business logic that invokes the Resolve method want to specify the amount of retries and the retryInterval? Those could also be given to the constructor of ArpMacResolverLwIp

lwip/lwip_cpp/MacResolverLwIp.hpp
```diff
@@ -0,0 +1,42 @@
...
+    {
+    public:
+        void SendRequest(IPv4Address address) override;
+        std::optional<hal::MacAddress> Lookup(IPv4Address address) override;
+        bool Resolve(IPv4Address address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone) override;
+
+    private:
+        void OnRetry();
+
+        IPv4Address pendingAddress_{};
```
richardapeters: I like to separate blocks of methods and member data by repeating the access specifier
```suggestion
    private:
        IPv4Address pendingAddress_{};
```

lwip/lwip_cpp/MacResolverLwIp.hpp
```diff
@@ -0,0 +1,42 @@
...
+    class Nd6MacResolverLwIp
+        : public Nd6MacResolver
+    {
+    public:
+        bool Resolve(const IPv6Address& address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone) override;
+
+    private:
+        void OnRetry();
+
+        IPv6Address pendingAddress_{};
```
richardapeters:
```suggestion
    private:
        IPv6Address pendingAddress_{};
```

lwip/lwip_cpp/MacResolverLwIp.hpp
richardapeters: The _ postfix for data members is not used in the rest of the code, please remove the _ postfix. If that causes clashes in implementation (assigning an onDone parameter to an onDone data member, we usually use this->onDone = onDone, or either of the two gets a more specific name (e.g. onResolvingDone)

lwip/lwip_cpp/MacResolverLwIp.hpp
```diff
@@ -0,0 +1,42 @@
...
+    public:
+        void SendRequest(IPv4Address address) override;
+        std::optional<hal::MacAddress> Lookup(IPv4Address address) override;
+        bool Resolve(IPv4Address address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone) override;
+
+    private:
+        void OnRetry();
+
+        IPv4Address pendingAddress_{};
+        infra::Function<void(std::optional<hal::MacAddress>)> onDone_;
```
richardapeters: After invoking a completion function, that function should be reset, since the lambda it contains may contain a SharedPtr or WeakPtr which needs to be released. Most often, AutoResetFunction does this trick best
```suggestion
        infra::AutoResetFunction<void(std::optional<hal::MacAddress>)> onDone_;
```

lwip/lwip_cpp/MacResolverLwIp.cpp
```diff
@@ -0,0 +1,178 @@
...
+            SendRequest(pendingAddress_);
+        }
+    }
+
+    bool Nd6MacResolverLwIp::Resolve(const IPv6Address& address, uint8_t retries, infra::Duration retryInterval, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone)
+    {
+        if (retryTimer_.Armed())
+        {
+            return false;
+        }
```
richardapeters: As this code suggests (and what I expect from the interface), Resolve may only be invoked again when the previous resolving has finished. Rather than silently ignore the next call if it's too early (probably breaking the logic of the object doing that call, which now waits for the result an infinite amount of time!), I would just abort on a second invocation before the previous one finished.

Would assertion on retryTimer_.Armed() be best? I'd rather choose onDone (on the premise that it's an AutoResetFunction)

```suggestion
        really_assert(!this->onDone);
```

lwip/lwip_cpp/MacResolverLwIp.cpp
```diff
@@ -0,0 +1,178 @@
...
+            auto* probe = pbuf_alloc_reference(nullptr, 0, PBUF_REF);
+            if (probe == nullptr)
+            {
+                return std::nullopt;
+            }
+
+            auto result = nd6_get_next_hop_addr_or_queue(netif_default, probe, &target, &hwaddrp);
+            pbuf_free(probe);
+
+            if (result >= 0 && hwaddrp != nullptr)
```
richardapeters:
```suggestion
            if (result == ERR_OK && hwaddrp != nullptr)
```

lwip/lwip_cpp/MacResolverLwIp.cpp
```diff
@@ -0,0 +1,178 @@
...
+        std::optional<hal::MacAddress> Nd6Lookup(const ip6_addr_t& target)
+        {
+            const u8_t* hwaddrp = nullptr;
+            auto* probe = pbuf_alloc_reference(nullptr, 0, PBUF_REF);
+            if (probe == nullptr)
+            {
+                return std::nullopt;
+            }
+
+            auto result = nd6_get_next_hop_addr_or_queue(netif_default, probe, &target, &hwaddrp);
```
richardapeters: But this doesn't actually lookup the target address, right? It merely finds the next hop address; if the target is not on the same network, then it finds the wrong MAC. Is there an implementation possible to find out if the target is on the local net, and return std::nullopt if it isn't?

services/network/MacResolver.hpp
richardapeters: Ah, I think I see from the implementation of Resolve that SendRequest and Lookup are there as helper functions for Resolve. In that case, they do not need to be part of the interface

uxitraHannes: Okay, I removed the Lookup and SendRequests form the interface and merge the interfaces. The reason for adding these methods in the interface was to make the usage more flexible. But the split is only possible for arp, so I think it is better with the simpler interface.

services/network/MacResolver.hpp
uxitraHannes: There is only on interface now, so yes: MacResolver is definitely the better name.

services/network/MacResolver.hpp
uxitraHannes: Moved the "configuration" to the constructor.

lwip/lwip_cpp/MacResolverLwIp.hpp
uxitraHannes: Fixed the suffix

lwip/lwip_cpp/MacResolverLwIp.cpp
uxitraHannes: Changed with using AutoResetFunction. Thank you for the finding!

lwip/lwip_cpp/MacResolverLwIp.cpp
uxitraHannes: Added a check for the used netif to check for "on-link". I don't see an other way of checking this.

lwip/lwip_cpp/MacResolverLwIp.cpp
uxitraHannes: Fixed

services/network/MacResolver.hpp
```diff
@@ -0,0 +1,27 @@
...
+{
+    class MacResolver
+    {
+    protected:
+        MacResolver() = default;
+        MacResolver(const MacResolver& other) = delete;
+        MacResolver& operator=(const MacResolver& other) = delete;
+
+    public:
+        virtual ~MacResolver() = default;
```
fabiangottstein:
```suggestion
    protected:
        MacResolver() = default;
        MacResolver(const MacResolver& other) = delete;
        MacResolver& operator=(const MacResolver& other) = delete;
        ~MacResolver() = default;

    public:
```

services/network/MacResolver.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+        using Address = std::variant<IPv4Address, IPv6Address>;
```
fabiangottstein: this alias already exists (in `services/network/Address.hpp`)

fabiangottstein: You can remove it here

lwip/lwip_cpp/MacResolverLwIp.hpp
```diff
@@ -0,0 +1,69 @@
...
+#define LWIP_MAC_RESOLVER_LW_IP_HPP
+
+#include "infra/timer/Timer.hpp"
+#include "infra/util/AutoResetFunction.hpp"
+#include "services/network/MacResolver.hpp"
+
+namespace services
+{
+    class MacResolverRetryHelper
+    {
```
fabiangottstein, line 11: Check out the `Retry` class, might be useful here.

lwip/lwip_cpp/MacResolverLwIp.hpp
```diff
@@ -0,0 +1,69 @@
...
+        infra::TimerRepeating retryTimer;
+        uint8_t retriesLeft = 0;
+    };
+
+    class ArpMacResolverLwIp
+        : public MacResolver
+    {
+    public:
+        ArpMacResolverLwIp(uint8_t retries, infra::Duration retryInterval);
+        ~ArpMacResolverLwIp() override = default;
```
fabiangottstein:
```suggestion
```

lwip/lwip_cpp/MacResolverLwIp.hpp
```diff
@@ -0,0 +1,69 @@
...
+    private:
+        MacResolverRetryHelper helper;
+    };
+
+    class Nd6MacResolverLwIp
+        : public MacResolver
+    {
+    public:
+        Nd6MacResolverLwIp(uint8_t retries, infra::Duration retryInterval);
+        ~Nd6MacResolverLwIp() override = default;
```
fabiangottstein:
```suggestion
```

lwip/lwip_cpp/MacResolverLwIp.cpp
```diff
@@ -0,0 +1,164 @@
+#include "lwip/lwip_cpp/MacResolverLwIp.hpp"
+#include "infra/util/ReallyAssert.hpp"
+#include "lwip/ip4_addr.h"
+#include "lwip/ip6_addr.h"
+#include "lwip/nd6.h"
+#include "lwip/netif.h"
+#include "lwip/pbuf.h"
+#include "netif/etharp.h"
```
fabiangottstein, line 8: is this include really needed?

lwip/lwip_cpp/MacResolverLwIp.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    void ArpMacResolverLwIp::Resolve(const Address& address, const infra::Function<void(std::optional<hal::MacAddress>)>& onResolveDone)
+    {
+        helper.Resolve(address, onResolveDone);
+    }
+
+    std::optional<hal::MacAddress> ArpMacResolverLwIp::Lookup(const Address& address) const
+    {
+        ip4_addr_t target;
+        auto& ipv4 = std::get<IPv4Address>(address);
```
fabiangottstein: you want to check the type before getting it ? Currently, it will abort if an IPv6 address is in there

lwip/lwip_cpp/MacResolverLwIp.cpp
uxitraHannes: Yes, symbols from all four headers are used

lwip/lwip_cpp/MacResolverLwIp.hpp
uxitraHannes: Used.

services/network/MacResolver.hpp
```diff
@@ -0,0 +1,24 @@
...
+    {
+    protected:
+        MacResolver() = default;
+        MacResolver(const MacResolver& other) = delete;
+        MacResolver& operator=(const MacResolver& other) = delete;
+        ~MacResolver() = default;
+
+    public:
+        virtual void Resolve(const IPAddress& address, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone) = 0;
+    };
```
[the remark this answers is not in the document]
uxitraHannes: Updated description

fabiangottstein (approved):

lwip/lwip_cpp/MacResolverLwIp.hpp
```diff
@@ -0,0 +1,67 @@
...
+    class MacResolverRetryHelper
+    {
+    public:
+        using LookupFunction = infra::Function<std::optional<hal::MacAddress>(const IPAddress&)>;
+        using SendRequestFunction = infra::Function<void(const IPAddress&)>;
+
+        MacResolverRetryHelper(uint8_t retries, infra::Duration retryInterval, LookupFunction lookup, SendRequestFunction sendRequest);
+
+        [[nodiscard]] bool Resolve(const IPAddress& address, const infra::Function<void(std::optional<hal::MacAddress>)>& onResolveDone);
+        [[nodiscard]] bool Busy() const;
```
richardapeters: Rather than trying to gracefully accommodate for the resolver already being busy, I would prefer that the responsibility lies with the consumer of this interface: That consumer should make sure that there's no invocation of Resolve() before the previous Resolve() has finished. That approach would be in line with how the other asynchronous interfaces in Embedded Infralib are designed. That would mean that Resolve() asserts that onDone is empty, instead of having to check Busy()

richardapeters (approved):

lwip/lwip_cpp/MacResolverLwIp.hpp
uxitraHannes: used really_assert on onDone, removed the Busy() method

lwip/lwip_cpp/MacResolverLwIp.hpp
```diff
@@ -0,0 +1,68 @@
...
+{
+    class MacResolverRetryHelper
+    {
+    public:
+        using LookupFunction = infra::Function<std::optional<hal::MacAddress>(const IPAddress&)>;
+        using SendRequestFunction = infra::Function<void(const IPAddress&)>;
+
+        MacResolverRetryHelper(uint8_t retries, infra::Duration retryInterval, LookupFunction lookup, SendRequestFunction sendRequest);
+
+        void Resolve(const IPAddress& address, const infra::Function<void(std::optional<hal::MacAddress>)>& onResolveDone);
```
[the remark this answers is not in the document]
uxitraHannes: Updated description

services/network/MacResolver.hpp
```diff
@@ -0,0 +1,24 @@
...
+    class MacResolver
+    {
+    protected:
+        MacResolver() = default;
+        MacResolver(const MacResolver& other) = delete;
+        MacResolver& operator=(const MacResolver& other) = delete;
+        ~MacResolver() = default;
+
+    public:
+        virtual void Resolve(const IPAddress& address, const infra::Function<void(std::optional<hal::MacAddress>)>& onDone) = 0;
```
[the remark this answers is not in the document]
uxitraHannes: Updated description

fabiangottstein:
@richardapeters the PR cannot be merged as the Sonar check is failing (is this an issue with the PR owner?)

richardapeters:
@fabiangottstein Ah, yes, that Sonar failure has recently been resolved on main. Can you merge in the current main? That should fix the failure.

uxitraHannes:
> Fabian Gottstein (@fabiangottstein) Ah, yes, that Sonar failure has recently been resolved on main. Can you merge in the current main? That should fix the failure.

The branch is up-to-date. The commit f68c29e950baa7460b34e5d49744d3f1af8b5aba (ci: sonarcloud build fixes ([#1191](https://github.com/uxitra/amp-embedded-infra-lib/issues/1191)) is applied.

I think there is a permission problem, maybe the job can be re-run with an internal user with access to the SonarQube project.

First error in job log:

"14:11:25.455 ERROR Error retrieving feature flags, using default analyzer behavior
Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator to check the permissions of the user the token belongs to"

PR #1186, opened 2026-05-01
feat: log and aborting with enum now supports integers

heinwessels-philips:
We have the option to `LOG_AND_ABORT_ENUM` which previously only compiled for enums. However, many times the C drivers we are using don't use enum types and instead store the enums as integer types. In these case we could not use this abort, even though it would make sense.

So this PR changes the `LOG_AND_ABORT_ENUM` to also support integer types. It doesn't compile if a different type is given than enum or integer. And it deduces the underlying value during compilation using some template magic.

### DIFF
--- a/infra/util/LogAndAbort.hpp
+++ b/infra/util/LogAndAbort.hpp
@@ -6,6 +6,24 @@
 #include <cstdlib>
 #include <type_traits>
 
+namespace
+{
+    template<class T, bool = std::is_enum_v<T>>
+    struct LogAndAbortEnumOrIntegralType
+    {
+        using Type = T;
+    };
+
+    template<class T>
+    struct LogAndAbortEnumOrIntegralType<T, true>
+    {
+        using Type = std::underlying_type_t<T>;
+    };
+
+    template<class T>
+    using LogAndAbortEnumOrIntegralTypeT = typename LogAndAbortEnumOrIntegralType<T>::Type;
+}
+
 namespace infra
 {
     using LogAndAbortHook = infra::Function<void(const char* reason, const char* file, int line, const char* format, va_list* args)>;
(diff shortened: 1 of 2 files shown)


infra/util/LogAndAbort.hpp
```diff
@@ -57,19 +82,19 @@ namespace infra
...
+        {                                                                                                                                                                                     \
+            LOG_AND_ABORT("Unexpected enum: %lld", static_cast<long long>(underlyingValue));                                                                                                  \
+        }                                                                                                                                                                                     \
+        else                                                                                                                                                                                  \
+        {                                                                                                                                                                                     \
+            LOG_AND_ABORT("Unexpected enum: %llu", static_cast<unsigned long long>(underlyingValue));                                                                                         \
```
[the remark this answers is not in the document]
heinwessels-philips: It would still be considered an enum.

infra/util/test/TestLogAndAbort.cpp
```diff
@@ -104,4 +105,24 @@ TEST_F(LogAndAbortTest, log_and_abort_enum)
...
+    int foo = 42;
+    EXPECT_DEATH(LOG_AND_ABORT_ENUM(foo), "");
+
+    uint64_t bar = 99;
+    EXPECT_DEATH(LOG_AND_ABORT_ENUM(bar), "");
+
+    uint64_t& baz = bar;
+    EXPECT_DEATH(LOG_AND_ABORT_ENUM(baz), "");
+
+    static_assert(infra::detail::LogAndAbortEnumOrIntegralType<std::decay_t<int>>::ToUnderlying(42) == int(42));
```
ccrugoPhilips, line 127: Just a small detail throughout this PR, please replace `integral` for `integer` :)

ccrugoPhilips (approved):

infra/util/test/TestLogAndAbort.cpp
heinwessels-philips: IMO `integral` is the right word. It comes from this: https://en.cppreference.com/cpp/types/is_integral

ccrugoPhilips: I find interesting they use integral in the standard... but also, from there:
image

It is not necessarily misleading, but I'd still argue that in embedded world we usually stick to `integer`. Up to you though 😃

heinwessels-philips: I'd prefer to keep it in line with the `is_integral`, since that's what I also use, which would be better for consistency.

PR #521, opened 2024-01-10
feat: changed EMIL_ENABLE_GLOBAL_TRACING define to EMIL_ENABLE_TRACING and EMIL_DISABLE_TRACING

daantimmer:

PR #1187, opened 2026-05-01
fix: disables tracer tests failing when compiled with EMIL_ENABLE_TRACING

heinwessels-philips:
## What failed:
When the disabled tracer tests was built with `EMIL_ENABLE_TRACING` then that would take precedence in the code. This would happen when EMIL was pulled into some external builds. And cause the tests to fail.

This is a regression after https://github.com/philips-software/amp-embedded-infra-lib/pull/521

## Reasons
It's because the CMake option `EMIL_ENABLE_TRACING` leaks into the code itself. There's two different things that needs to happen, which was combined into a single variable
- CMake should tell the tracer library to enable tracing
- The library should change the code accodingly

This causes issues when not used correctly:
- For example parent library setting `compile_time_target` instead of `set(EMIL_ENABLE_TRACING On)`. This will currently result in unexpected behaviour, but with this fix it will do nothing.
- Some code including the `Tracer.hpp` header without linking the library. Previously this would be unexpected behaviour, but with this fix it won't compile.

## Fix

This PR separates these two things, which should resolve the issue.

It will still conform to the requirement of:
- If `EMIL_ENABLE_TRACING` is defined at all, then it will take precedence and the tracer will be enabled

### DIFF
--- a/services/tracer/Tracer.hpp
+++ b/services/tracer/Tracer.hpp
@@ -15,10 +15,10 @@ namespace services
         Tracer& operator=(const Tracer& other) = delete;
         virtual ~Tracer() = default;
 
-#if defined(EMIL_ENABLE_TRACING)
+#if defined(EMBEDDED_INFRA_SERVICES_TRACER_ENABLED)
         infra::TextOutputStream Trace();
         virtual infra::TextOutputStream Continue() = 0;
-#elif defined(EMIL_DISABLE_TRACING)
+#elif defined(EMBEDDED_INFRA_SERVICES_TRACER_DISABLED)
         class EmptyTracing
         {
         public:
@@ -65,7 +65,7 @@ namespace services
     public:
         explicit TracerToStream(infra::TextOutputStream& stream);
 
-#if defined(EMIL_ENABLE_TRACING)
+#if defined(EMBEDDED_INFRA_SERVICES_TRACER_ENABLED)
         infra::TextOutputStream Continue() override;
 #endif
 
@@ -79,7 +79,7 @@ namespace services
     public:
         explicit TracerToDelegate(Tracer& delegate);
 
-#if defined(EMIL_ENABLE_TRACING)
+#if defined(EMBEDDED_INFRA_SERVICES_TRACER_ENABLED)
         infra::TextOutputStream Continue() override;
 #endif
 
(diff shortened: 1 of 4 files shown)


services/tracer/Tracer.hpp
```diff
@@ -15,10 +15,10 @@ namespace services
         Tracer& operator=(const Tracer& other) = delete;
         virtual ~Tracer() = default;
 
-#if defined(EMIL_ENABLE_TRACING)
+#if defined(EMBEDDED_INFRA_SERVICES_TRACER_ENABLED)
```
andjordan, line 18: I think renaming it hides the violation from the upper repository
`EMIL_ENABLE_TRACING` shouldn't take precedence over `EMIL_DISABLE_TRACING` and having both defined and equal to 1 should result in a compile error, so the upper repository can fix it correctly (similar as not having any of them defined, as already handled in #else)

heinwessels-philips: > I think renaming it hides the violation from the upper repository

Fair, but with new code the violation will result in nothing happening instead of it breaking in weird ways, which is better IMO.

> EMIL_ENABLE_TRACING shouldn't take precedence over EMIL_DISABLE_TRACING
That's how it was before though?

> having both defined and equal to 1 should result in a compile error
This can easily still be handled in CMake
```cmake
if (EMIL_ENABLE_TRACING && EMIL_DISABLE_TRACING )
    messages(FATAL_ERROR "Cannot set both EMIL_ENABLE_TRACING  and EMIL_DISABLE_TRACING ")
elseif(EMIL_ENABLE_TRACING)
    target_compile_definitions(services.tracer PUBLIC EMBEDDED_INFRA_SERVICES_TRACER_ENABLED=1)
else()
    target_compile_definitions(services.tracer PUBLIC EMBEDDED_INFRA_SERVICES_TRACER_DISABLED=1)
endif()
```

I think this change should still be considered to clean this interface for the reasons described in the PR description.

> - For example parent library setting `compile_time_target` instead of `set(EMIL_ENABLE_TRACING On)`. This will currently result in unexpected behaviour, but with this fix it will do nothing.
> - Some code including the `Tracer.hpp` header without linking the library. Previously this would be unexpected behaviour, but with this fix it won't compile.

heinwessels-philips:
After chat with Andre it was decided that the benefit is too minor to break the CCache

PR #1195, opened 2026-05-11
feat: add EcSecP256r1PublicKey

richardapeters:
## Pull request overview

This PR extends the Sesame Mbed TLS crypto utilities with support for exporting/consuming an EC secp256r1 public key in DER form, and adds unit tests to cover signing/verifying both with raw public keys and with certificates.

**Changes:**
- Added `services::EcSecP256r1PublicKey` for deriving/exporting a secp256r1 public key (DER) from a private key.
- Added an `EcSecP256r1DsaVerifierMbedTls` constructor overload that accepts a DER-encoded public key directly.
- Added Mbed TLS–gated unit tests for sign/verify flows using both a raw public key and a certificate chain.

### DIFF
--- a/services/util/SesameCryptoMbedTls.hpp
+++ b/services/util/SesameCryptoMbedTls.hpp
@@ -49,13 +49,13 @@ namespace services
         : public EcSecP256r1DsaVerifier
     {
     public:
+        EcSecP256r1DsaVerifierMbedTls(infra::ConstByteRange dsaPublicKey);
         EcSecP256r1DsaVerifierMbedTls(infra::ConstByteRange dsaCertificate, infra::ConstByteRange rootCaCertificate);
         ~EcSecP256r1DsaVerifierMbedTls();
 
         bool Verify(infra::ConstByteRange data, infra::ConstByteRange r, infra::ConstByteRange s) const override;
 
     private:
-        mbedtls_x509_crt rootCertificate;
         mbedtls_ecp_group group;
         mbedtls_ecp_point publicKey;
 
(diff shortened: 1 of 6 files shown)


services/util/SesameCryptoMbedTls.hpp
```diff
@@ -119,6 +119,20 @@ namespace services
...
+
+    class EcSecP256r1PublicKey
+    {
+    public:
+        EcSecP256r1PublicKey(const EcSecP256r1PrivateKey& privateKey, hal::SynchronousRandomDataGenerator& randomDataGenerator);
+        EcSecP256r1PublicKey(const EcSecP256r1PublicKey& other) = delete;
+        EcSecP256r1PublicKey& operator=(const EcSecP256r1PublicKey& other) = delete;
+        ~EcSecP256r1PublicKey();
+
+        std::array<uint8_t, 91> Der() const;
```
andjordan: Is the 91 available in a constant? If not, maybe `using DerType = std::array<uint8_t, 91>` (with a better name :))

andjordan (approved):

services/util/SesameCryptoMbedTls.hpp
richardapeters: No, but I'll make a type definition for the array that contains the DER-encoded private/public keys

fabiangottstein:
@richardapeters Those services are not necessarily limited to Sesame usage, would it make sense to rename the source files to encourage more broader usage?

richardapeters:
@fabiangottstein Eventually, yes, but I'd like to first build up more experience with the interface

PR #1196, opened 2026-05-12
feat!: expand GapCentral ECHO interface

heinwessels-philips:
Consolidates the GapCentral interface and adds a bit more functionality to it. These are breaking changes.

- In `service GapCentral` may `method_id`s changed.
- Removed support for managing `Out-Of-Band` data
- Removed `GapPeripheralReadyness::DeviceStarted` from Central, and deprecate for peripheral. This is replaced with a `Gap*.CurrentState(GapState::Standby)`.
- Rename `PairingStatus` to `PairingFailedReason`, and expand with more reasons to better reflect the spec.
- Marked `IsDeviceBonded` as deprecated to smooth transision.


services/ble/Gap.proto
```diff
@@ -142,16 +152,27 @@ message BondsStatus
...
 }
 
 message AdvertisementMode
 {
     AdvertisementType type = 1;
-    uint32 advInterval = 2;
+
+    // In increments of 0.625 ms, as per Bluetooth Core Specification.
+    // This will be used for both minimum and maximum advertising interval.
+    // TODO: Explain what min/max is. Or link to standard
```
oguzcanoguz: Did we include all concrete todos here and all open comments in the story: https://tfsemea1.ta.philips.com/tfs/TPC_Region14/Connectivity/_workitems/edit/179081

Currently in the story I don't see remnant comments.

services/ble/Gap.proto
```diff
@@ -326,112 +369,228 @@ message SecurityModeAndLevel
...
+    // During pairing, GapCentralResponse.AuthenticationRequired may be called if user interaction is required.
+    // Results in GapCentralResponse.PairingSuccess or GapCentralResponse.PairingFailed
+    // Can result in GapCentralResponse.NumberOfBondsChanged or GapCentralResponse.CurrentState with standby if device disconnected
     rpc PairAndBond(Nothing) returns (Nothing) { option (method_id) = 6; }
 
+    // Allowed in link layer states: standby
+    // Configures the minimum security mode and level required for connections.
+    // Should be called once during initialization before any connections are made.
+    // If a connected peer does not meet the required security level, the connection will be rejected.
     rpc SetSecurityMode(SecurityModeAndLevel) returns (Nothing) { option (method_id) = 7; }
```
oguzcanoguz, line 519: I forgot to mention this earlier, but we want to modify this part as it is done in peripheral. Hence using securitypreference. That was a clarification we did on GapPeripheral. 
Please add as a todo.

services/ble/Gap.proto
```diff
@@ -326,112 +369,228 @@ message SecurityModeAndLevel
...
+        // Filters on the device address (public or random). Passes if the address matches exactly.
+        // data: the device address to match (see Address).
+        address = 1;
 
-        // Filtering on contents of Manufacturer specific data
-        advertisingData = 1; 
+        // Filters on the full contents of the Manufacturer Specific Data field in the advertising or scan response data.
+        // Passes if the manufacturer specific data matches exactly.
+        // data: the manufacturer specific data to match, including the 2-byte manufacturer ID.
+        manufacturerSpecificData = 2;
```
oguzcanoguz, line 428: We had an open comment to extend this filter list I think.

services/ble/Gap.proto
```diff
@@ -326,112 +369,228 @@ message SecurityModeAndLevel
...
     // Allowed states: connected
-    // 1. If there is a pre-existing bond, then the connection will be encrypted.
-    // 2. If there is no pre-existing bond, then pairing, encrypting, and bonding (storing the keys) will take place.
+    // Can be called to secure a existing connection.
+    // If the peripheral sends a security request, this procedure is triggered automatically.
+    // 1. If there is no pre-existing bond, then pairing, encrypting, and bonding (storing the keys) will take place.
+    // 2. If there is a pre-existing bond, then the connection will be encrypted using the existing keys.
+    // During pairing, GapCentralResponse.AuthenticationRequired may be called if user interaction is required.
+    // Results in GapCentralResponse.PairingSuccess or GapCentralResponse.PairingFailed
+    // Can result in GapCentralResponse.NumberOfBondsChanged or GapCentralResponse.CurrentState with standby if device disconnected
```
oguzcanoguz: I'd split this line into two and add if NumberOfBondsChanged  will be received if there is a new bond created.

services/ble/Gap.proto
```diff
@@ -326,112 +369,228 @@ message SecurityModeAndLevel
...
+
+    // Allowed in link layer states: standby
+    // Sets the local identity address used by the central when privacy mode is disabled.
+    // The address type may be public or random.
+    // This does not set a resolvable private address; when privacy mode is enabled, the controller generates and rotates resolvable private addresses automatically.
+    rpc SetIdentityAddress(AddressWithType) returns (Nothing) { option (method_id) = 15; }
+
+    // Allowed in link layer states: standby
+    // Enables or disables use of a resolvable private address for the central.
+    // When enabled, the controller generates and rotates the private address automatically based on the local IRK. It will reset any set identity address.
```
oguzcanoguz: When privacy is turned on, the device still has an identity address which should be the same with the set identity address. That address still determines device's identity and communicated to the peer device during bonding time.

services/ble/Gap.proto
```diff
@@ -326,112 +369,228 @@ message SecurityModeAndLevel
...
+    // When disabled, the central uses its configured identity address.
+    rpc SetPrivacyMode(BoolValue) returns (Nothing) { option (method_id) = 16; }
+
+    // DEPRECATED: Do not use this for new code. It will be removed in the near future.
+    // Responds with GapCentralResponse.DeviceBonded
+    rpc IsDeviceBonded(AddressWithType) returns (Nothing) { option (method_id) = 17; }
```
oguzcanoguz, line 577: They still have to use it until we implement new bonding features.

oguzcanoguz (approved):

services/ble/Gap.proto
heinwessels-philips: I'll change this now

services/ble/Gap.proto
heinwessels-philips: This TODO i did resolve, seems I didn't push it.
```.proto
    // Valid range: 0x0020 to 0x4000 (20 ms to 10.24 s).
    // See Bluetooth Core Specification v6.3, Vol 4, Part E, Section 7.8.5.
```

All other TODOs I already resolved.

services/ble/Gap.proto
heinwessels-philips: This comment is still here. Should I added the ServiceUUID filter already? My plan was to wait until they can tell us which filter they need.
https://github.com/philips-software/amp-embedded-infra-lib/blob/91043cdc53b17596093ab4347638812cb802689a/services/ble/Gap.proto#L430

services/ble/Gap.proto
heinwessels-philips: Actually, I don't have enough information to do this. I'll add a TODO.

services/ble/Gap.proto
```diff
@@ -286,16 +314,14 @@ service GapPeripheralResponse
...
-
+    rpc NumberOfBondsChanged(UInt32Value) returns(Nothing) { option(method_id) = 6; }
 
-    rpc NumberOfBondsChanged(UInt32Value) returns(Nothing) { option(method_id) = 7; }
-
-    rpc BondInformation(BondsStatus) returns(Nothing) { option(method_id) = 8; }
+    rpc BondInformation(BondsStatus) returns(Nothing) { option(method_id) = 7; }
 }
 
+// DEPRECATED: Will be removed in the near future. Instead wait for CurrentState(standby) after startup
```
EkelmansPh, line 323: Why not remove it now? This proto already contains a number of other deprecated things/breaking changes.

heinwessels-philips: To reduce scope of this task. If we remove this we more upstream code that needs to be updated. So this is for our benefit. It will likely be removed in Q3.

services/ble/Gap.proto
```diff
@@ -326,112 +369,231 @@ message SecurityModeAndLevel
...
+    uint32 initiatingTimeoutInMs = 2;
 }
 
-message ConnectionParameters
+message AdvertisingReportType
 {
-    PeerNodeParameters peerParameters = 1;
-    uint32 initiatingTimeoutInMs = 2;
+    // Types defined in Bluetooth Core Specification v6.3, Vol 4, Part E, Section 7.7.65.2
+    enum AdvertisingReportTypeEnum
```
EkelmansPh, line 387: aren't these the same types as found in 'AdvertisementTypeEnum'? If so, could they be reused?

services/ble/test_doubles/GapPairingObserverMock.hpp
```diff
@@ -12,9 +12,8 @@ namespace services
     public:
         using GapPairingObserver::GapPairingObserver;
 
-        MOCK_METHOD(void, DisplayPasskey, (int32_t passkey, bool numericComparison), (override));
-        MOCK_METHOD(void, PairingSuccessfullyCompleted, (), (override));
-        MOCK_METHOD(void, PairingFailed, (PairingErrorType error), (override));
+        MOCK_METHOD(void, AuthenticationRequired, (bool numericComparison, uint32_t passkey), (override));
+        MOCK_METHOD(void, PairingResult, (bool pairedSuccessfully, PairingFailedReason pairingFailedReason), (override));
         MOCK_METHOD(void, OutOfBandDataGenerated, (const GapOutOfBandData& outOfBandData), (override));
```
EkelmansPh, line 17: Shouldn't this be removed? or was OOB only removed from the proto file?

services/ble/Gap.proto
heinwessels-philips: Good catch. There's a difference between an AdvertisingReport (received by Central) and AdvertisingType (used by peripheral). One could have additional types like `ScanResponse`, which doesn't make sense for peripheral, so it was decided to keep it separate.

services/ble/test_doubles/GapPairingObserverMock.hpp
heinwessels-philips: It's only removed from the interface, but still supported by our drivers.

services/ble/Gap.hpp
```diff
@@ -85,6 +92,12 @@ namespace services
         infra::ConstByteRange confirmData;
     };
 
+    struct Bond
+    {
+        GapAddress address;
+        infra::BoundedConstString deviceName;
+        bool isCurrentlyConnected;
```
andjordan: I think this is not a bonding information. Isn't there another interface to report the connected device?

services/ble/Gap.hpp
```diff
@@ -226,9 +250,12 @@ namespace services
         // Implementation of GapBonding
         void RemoveAllBonds() override;
         void RemoveOldestBond() override;
+        void RemoveBondWithAddress(GapAddress gapAddress) override;
+
         std::size_t GetMaxNumberOfBonds() const override;
         std::size_t GetNumberOfBonds() const override;
         bool IsDeviceBonded(hal::MacAddress address, GapDeviceAddressType addressType) const override;
+        std::pair<infra::MemoryRange<const services::Bond>, uint32_t> GetBondList() const override;
```
andjordan: what is the uint32_t here?

services/ble/Gap.proto
```diff
@@ -64,72 +65,87 @@ message SecurityLevel
...
     {
         publicAddress = 0;
+
+        // Random addresses can be either static random or private (resolvable or non-resolvable).
+        // The subtype is encoded in the two most significant bits of the address (bits 47-46):
+        //   00 - Non-resolvable private: rotated periodically, cannot be resolved.
+        //   01 - Resolvable private (RPA): rotated periodically, can be resolved using the peer's IRK (Identity Resolving Key).
+        //   11 - Static random: fixed for the lifetime of the device or until a power cycle.
+        // See Bluetooth Core Specification v6.3, Vol 6, Part B, Section 1.3.2.
         randomAddress = 1;
```
andjordan, line 90: shouldn't the enum list all possible types, as the comment is describing all of them?

services/ble/Gap.proto
```diff
@@ -64,72 +65,87 @@ message SecurityLevel
...
-        pairingNotSupported = 2;
-        insufficientEncryptionKeySize = 3;
-        numericComparisonFailed = 4;
-        timeout = 5;
-        encryptionFailed = 6;
-        unknown = 7;
-    }
-
-    Error error = 1;
+    bool pairedSuccessfully = 1;    // If false then 'reason' will indicate the failure reason.
```
andjordan, line 146: Is there a particular reason for changing from 2 method to 1 with a boolean flag?

services/ble/Gap.proto
```diff
@@ -326,112 +369,230 @@ message SecurityModeAndLevel
...
-    bytes confirmData = 4 [(bytes_size) = 16];
+    bool success = 1; // True if the address was successfully resolved to a bond, false otherwise.
+    Address identityAddress = 2; // The resolved identity address
+}
+
+message Bond
+{
+    AddressWithType addressWithType = 1;        // Identity address
+    bytes deviceName = 2 [(bytes_size) = 32];   // Device Name from GAP (Generic Access) service. Not unique; truncated to 32 bytes.
+    bool isCurrentlyConnected = 3;              // Is this device currently connected.
```
andjordan: same as Gap.hpp

services/ble/Gap.proto
```diff
@@ -326,112 +369,230 @@ message SecurityModeAndLevel
...
+    // Resolves a resolvable private address to the identity address of an already bonded device.
+    // Useful to identify whether a device seen during scanning is already bonded, without needing to connect first.
+    // Will only succeed if the address is a resolvable private address and the device is bonded.
+    // Results in GapCentralResponse.ResolvePrivateAddressResult
     rpc ResolvePrivateAddress(Address) returns (Nothing) { option (method_id) = 13; }
 
-    // This will work with both public and random addresses
-    rpc IsDeviceBonded(PeerNodeParameters) returns (Nothing) { option (method_id) = 14; }
-    rpc GenerateOutOfBandData(Nothing) returns (Nothing) { option (method_id) = 15; }
-    rpc SetOutOfBandData(OutOfBandData) returns (Nothing) { option (method_id) = 16; }
```
andjordan, line 415: aren't other users of the library using this feature?

heinwessels-philips: We asked around and nobody is using it. There's only people using the HAL ST implementation, but nobody uses it over Echo.

services/ble/Gap.proto
heinwessels-philips: This was asked by a client, and makes the API easier to interact with. It allows your code to be more localised.

services/ble/Gap.proto
heinwessels-philips: Yeah probably. I think because of the ST implementation the correct name for this should rather be `IdentityAddressType`. Or as you say we should make the enum more explicit. But considered out of scope for now, since I didn't touch the types in the PR. Making a note in an upcoming task.

heinwessels-philips: I think originally this was supposed to be called `IdentityAddressType`. But I do agree, this is inconsistent. Either needs to be renamed, or expanded with all the types. I didn't touch this in this PR though, so seeing this as out of scope. Made a note to look into this in the Q3 task.

services/ble/Gap.hpp
heinwessels-philips: The `uint32_t` is the `maximumAllowedBonds` that comes with `BondList`

services/ble/Gap.hpp
heinwessels-philips: Yeah, you're right. There is no other way to know which device is currently connected. I think I'll remove this rather, and then we can see in Q3 if we want to address this, because nobody requested this.

services/ble/Gap.hpp
andjordan: Isn't that the same as `GetMaxNumberOfBonds()` ?

heinwessels-philips: `GetMaxNumberOfBonds` no longer exists

services/ble/Gap.hpp
```diff
@@ -370,11 +395,9 @@ namespace services
     {
     public:
         virtual void Connect(hal::MacAddress macAddress, GapDeviceAddressType addressType, infra::Duration initiatingTimeout) = 0;
-        virtual void CancelConnect() = 0;
-        virtual void Disconnect() = 0;
+        virtual void Standby() = 0;
         virtual void SetIdentityAddress(hal::MacAddress macAddress, GapDeviceAddressType addressType) = 0;
         virtual void StartDeviceDiscovery() = 0;
-        virtual void StopDeviceDiscovery() = 0;
```
EkelmansPh, line 377: Why no more stop discovery?

heinwessels-philips: It's replaced with `Standby()`, similar to peripheral's advertising is stopped with `Standby`

services/ble/Gap.hpp
EkelmansPh: Sounds like GetBondList is now overloaded if it returns both the list of bonds and the maximum number of bonds.

EkelmansPh (approved):

services/ble/Gap.hpp
heinwessels-philips: I agree. Removed `maximumAllowedBonds`, to be revisited in Q3

services/ble/Gap.proto
```diff
@@ -326,112 +368,228 @@ message SecurityModeAndLevel
...
-message OutOfBandData
+message ResolvedPrivateAddress
 {
-    Address address = 1;
-    AddressType addressType = 2;
-    bytes randomData = 3 [(bytes_size) = 16];
-    bytes confirmData = 4 [(bytes_size) = 16];
+    bool success = 1; // True if the address was successfully resolved to a bond, false otherwise.
+    Address identityAddress = 2; // The resolved identity address
+}
```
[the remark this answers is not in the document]
heinwessels-philips: Made a note to revisit in Q3

PR #1206, opened 2026-05-19
feat: prefixed tracer

andjordan:
This enables the application to provide a tracer with a prefix and distinguish between different sources.

For example, on some applications there might be more than one echo connections, and the user can use TracerPrefixed to distinguish between multiple connections

### DIFF
--- a/services/tracer/Tracer.hpp
+++ b/services/tracer/Tracer.hpp
@@ -128,6 +128,26 @@ namespace services
     private:
         uint8_t colour;
     };
+
+    class TracerPrefixed
+        : public TracerToDelegate
+    {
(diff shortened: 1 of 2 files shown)


services/tracer/Tracer.hpp
```diff
@@ -128,6 +128,26 @@ namespace services
     private:
         uint8_t colour;
     };
+
+    class TracerPrefixed
```
heinwessels-philips, line 132: Maybe?
```suggestion
    class TracerPrefixedToDelegate
```

heinwessels-philips (approved):
This is a very cool idea!

PR #1207, opened 2026-05-19
chore: wait for read/write done before continuing with the operation

andjordan:


services/util/FlashEcho.cpp
```diff
@@ -125,13 +125,15 @@ namespace services
         bufferPosition = 0;
         writingBuffer = buffer;
         this->onDone = onDone;
+        this->address = address;
```
heinwessels-philips, line 128: Could a unit test be added for this behaviour?

heinwessels-philips (changes requested):

services/util/FlashEcho.cpp
andjordan: PR is not ready yet.

PR #1211, opened 2026-05-20
feat: sequential requests on flash over echo

andjordan:
This PR adds support for choosing between *pipelined* vs *sequential* chunked flash read/write behavior when using the Echo-based flash proxy, and expands the unit tests to validate both behaviors.

**Changes:**
- Refactors the existing `FlashEchoProxy` into a shared `FlashEchoProxyBase`, and introduces `FlashEchoSequentialProxy` for sequential chunk dispatch.
- Adds new unit tests to verify that `FlashEchoProxy` pipelines chunk requests while `FlashEchoSequentialProxy` waits for each `*Done` before issuing the next chunk.
- Updates homogeneous proxy construction to work with the refactor.

### DIFF
--- a/services/util/FlashEcho.cpp
+++ b/services/util/FlashEcho.cpp
@@ -182,16 +182,16 @@ namespace services
         MethodDone();
     }
 
-    void FlashEchoProxyBase::OnReadIncomplete(uint32_t, uint32_t)
+    void FlashEchoProxyBase::OnReadIncomplete([[maybe_unused]] uint32_t address, [[maybe_unused]] uint32_t bufferPosition)
     {}
 
-    void FlashEchoProxyBase::OnWriteIncomplete(uint32_t, uint32_t)
+    void FlashEchoProxyBase::OnWriteIncomplete([[maybe_unused]] uint32_t address, [[maybe_unused]] uint32_t bufferPosition)
     {}
 
-    void FlashEchoProxyBase::OnReadChunkSent(uint32_t, uint32_t)
+    void FlashEchoProxyBase::OnReadChunkSent([[maybe_unused]] uint32_t address, [[maybe_unused]] uint32_t nextStart)
     {}
 
-    void FlashEchoProxyBase::OnWriteChunkSent(uint32_t, uint32_t)
+    void FlashEchoProxyBase::OnWriteChunkSent([[maybe_unused]] uint32_t address, [[maybe_unused]] uint32_t nextStart)
     {}
 
     void FlashEchoProxyBase::ReadPartialBuffer(uint32_t address, uint32_t start)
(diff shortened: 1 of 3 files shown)


services/util/test/TestFlashEcho.cpp
```diff
@@ -197,3 +197,263 @@ TEST_F(FlashProxyTest, EraseSectors)
...
+                            flashResult.ReadDone(infra::Head(infra::MakeRange(chunk), size));
+                        });
+                };
+            });
+
+    infra::VerifyingFunction<void()> onDone;
+    flashProxy.ReadBuffer(infra::MakeRange(buffer), 1234, onDone);
+
+    // Both reads were requested before any ReadDone (pipelined)
+    ASSERT_TRUE(sendReadDone1);
```
richardapeters: ASSERT -> EXPECT

services/util/FlashEcho.hpp
```diff
@@ -46,24 +46,57 @@ namespace services
...
+
     private:
         // Implementation of flash::FlashResult
         void ReadDone(infra::ConstByteRange contents) override;
         void WriteDone() override;
         void EraseSectorsDone() override;
 
-        void ReadPartialBuffer(uint32_t address, uint32_t start);
-        void WritePartialBuffer(uint32_t address, uint32_t start);
+        virtual void OnReadIncomplete() = 0;
```
richardapeters: I would give these 4 function empty implementations, so that in the derived classes you can override 2 of them, without bothering with the other 2

richardapeters: I would add address and bufferStart as parameters to OnReadIncomplete and OnWriteIncomplete, so that the data members can be made private

services/util/FlashEcho.hpp
```diff
@@ -30,12 +30,12 @@ namespace services
...
     };
 
-    class FlashEchoProxy
+    class FlashEchoProxyBase
         : public hal::Flash
         , private flash::FlashResult
     {
     public:
-        FlashEchoProxy(services::Echo& echo, infra::MemoryRange<const uint32_t> sectorSizes);
+        FlashEchoProxyBase(services::Echo& echo, infra::MemoryRange<const uint32_t> sectorSizes);
```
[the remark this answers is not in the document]
andjordan: not deleted from the base class interface

services/util/FlashEcho.cpp
```diff
@@ -149,34 +151,50 @@ namespace services
...
 
-    void FlashEchoProxy::EraseSectorsDone()
+    void FlashEchoProxyBase::EraseSectorsDone()
     {
         onDone();
         MethodDone();
     }
 
-    void FlashEchoProxy::ReadPartialBuffer(uint32_t address, uint32_t start)
+    void FlashEchoProxyBase::OnReadIncomplete(uint32_t, uint32_t)
```
richardapeters: Where are the parameter names :)

services/util/FlashEcho.cpp
```diff
@@ -199,13 +217,33 @@ namespace services
...
+    {
+        ReadPartialBuffer(address, nextStart);
+    }
+
+    void FlashEchoProxy::OnWriteChunkSent(uint32_t address, uint32_t nextStart)
+    {
+        WritePartialBuffer(address, nextStart);
+    }
+
+    void FlashEchoSequentialProxy::OnReadIncomplete(uint32_t address, uint32_t bufferPosition)
```
richardapeters: is "bufferPosition" the right name?

richardapeters (approved):

services/util/FlashEcho.cpp
andjordan: it is either this or [[maybe_unused]] uint32_t address

services/util/FlashEcho.cpp
andjordan: nope, aligned with the other  method

PR #1212, opened 2026-05-20
feat: flash echo homogeneous sequential proxy

andjordan:
This PR extends the existing Flash Echo proxy utilities by generalizing the “homogeneous flash” proxy so it can be used with both the pipelined and sequential FlashEcho proxy behaviors, and slightly adjusts Sesame echo initialization ordering.

**Changes:**
- Refactor `FlashEchoHomogeneousProxy` into a templated `FlashEchoHomogeneousProxyBase<T>` and add a new `FlashEchoHomogeneousSequentialProxy` alias.
- Implement the templated homogeneous proxy methods in `FlashEcho.cpp` and explicitly instantiate for `FlashEchoProxy` and `FlashEchoSequentialProxy`.


services/util/EchoOnSesame.cpp
andjordan, line 1: Ignore this file (to be reverted before merge)

services/util/EchoOnSesame.cpp
```diff
@@ -25,7 +25,7 @@ namespace services
         initialized = true;
 
         EchoOnStreams::Initialized();
-
+
```
andjordan:
```suggestion
```

services/util/EchoOnSesame.cpp
```diff
@@ -25,7 +25,6 @@ namespace services
         initialized = true;
 
         EchoOnStreams::Initialized();
```
andjordan, line 27:
```suggestion
        EchoOnStreams::Initialized();

```

richardapeters (approved):

PR #1223, opened 2026-05-28
chore: add set privacy mode to gap central interface

gabrielsantosphilips:
add set privacy mode to gap central interface.

### DIFF
--- a/services/ble/Gap.proto
+++ b/services/ble/Gap.proto
@@ -413,7 +413,7 @@ service GapCentral
     rpc IsDeviceBonded(PeerNodeParameters) returns (Nothing) { option (method_id) = 14; }
     rpc GenerateOutOfBandData(Nothing) returns (Nothing) { option (method_id) = 15; }
     rpc SetOutOfBandData(OutOfBandData) returns (Nothing) { option (method_id) = 16; }
-    rpc SetAddress(AddressWithType) returns (Nothing) { option (method_id) = 17; }
+    rpc SetIdentityAddress(AddressWithType) returns (Nothing) { option (method_id) = 17; }
     rpc SetPrivacyMode(BoolValue) returns (Nothing) { option (method_id) = 18; }
 }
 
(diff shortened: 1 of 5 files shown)


services/ble/Gap.proto
```diff
@@ -413,6 +413,8 @@ service GapCentral
     rpc IsDeviceBonded(PeerNodeParameters) returns (Nothing) { option (method_id) = 14; }
     rpc GenerateOutOfBandData(Nothing) returns (Nothing) { option (method_id) = 15; }
     rpc SetOutOfBandData(OutOfBandData) returns (Nothing) { option (method_id) = 16; }
+    rpc SetAddress(AddressWithType) returns (Nothing) { option (method_id) = 17; }
```
heinwessels-philips: If I understand correctly we cannot set the private address? So we can more explicit for what this function is used for.
```suggestion
    rpc SetIdentityAddress(AddressWithType) returns (Nothing) { option (method_id) = 17; }
```

services/ble/Gap.proto
```diff
@@ -413,6 +413,8 @@ service GapCentral
...
+    rpc SetPrivacyMode(BoolValue) returns (Nothing) { option (method_id) = 18; }
```
heinwessels-philips, line 417: Would we want to add parameters here, like for example the rotate period? According to copilot the default is 15 minutes. I guess that's easy to add later, so this is more of a nitpick thought

heinwessels-philips (changes requested):

services/ble/Gap.proto
gabrielsantosphilips: I think there is no request to change that period (however I like the idea) and adding this we need somehow verify if the RPA is rotating, which is out of scope.

services/ble/Gap.proto
gabrielsantosphilips: I agree, name updated

heinwessels-philips (approved):

oguzcanoguz (approved):

PR #1222, opened 2026-05-28
docs: add information about event dispatcher

heinwessels-philips:
This adds more information about the EventDispatcher, including common pitfalls and work arounds. This is for developers to read, and also importantly, for AI reviewers to read when touching code with EventDispatchers.

PR #1227, opened 2026-06-01
chore: add copilot instructions for review

heinwessels-philips:
Adds instructions for Co-Pilot to improve the reviews in this repository. The main goal is to include notes about using the EventDispatcher, as added in #1222. Also updates the coding standard to nto use `final`.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -33,8 +33,8 @@ if (TARGET_MCU_VENDOR STREQUAL ti)
 cmake --list-presets
 
 # Configure and build for target
-cmake --preset <configure_preset>
-cmake --build --preset <build_preset>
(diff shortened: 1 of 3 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,90 @@
+# Copilot Instructions for amp-embedded-infra-lib
```
andjordan, line 1: can you add an instruction that tests with `EXPECT_DEATH` must be guarded with `#ifndef EMIL_MUTATION_TESTING`?
Mutation tests with EXPECT_DEATH is very slow and can crash the runner: https://github.com/philips-software/amp-embedded-infra-lib/actions/runs/26746280655/job/78822476173

BarisTanyeri (approved):

PR #1237, opened 2026-06-10
fix: fix sesame received decryption Finish outputs leftover after Update

dcayy:
In SesameSecured message received, Finish() can still output leftover that were not returned by Update().


services/util/SesameSecured.cpp
```diff
@@ -141,7 +141,9 @@ namespace services
         }
 
         std::array<uint8_t, blockSize> computedMac;
-        receiveEncryption.Finish(infra::ByteRange(), computedMac);
+        receiveBuffer.resize(receiveBuffer.size() + blockSize);
+        auto moreProcessedSize = receiveEncryption.Finish(infra::Tail(infra::MakeRange(receiveBuffer), blockSize), computedMac);
+        receiveBuffer.resize(receiveBuffer.size() - blockSize + moreProcessedSize);
```
heinwessels-philips: Would it be possible to add a unit test to verify this behavior and prevent regressions?

dcayy: done

services/util/SesameSecured.cpp
```diff
@@ -141,7 +141,9 @@ namespace services
         }
 
         std::array<uint8_t, blockSize> computedMac;
-        receiveEncryption.Finish(infra::ByteRange(), computedMac);
+        receiveBuffer.resize(receiveBuffer.size() + blockSize);
```
richardapeters: Please resize to the exact size needed (total encrypted size minus total decrypted size until now), and check that exactly that size has been produced by Finish

richardapeters (approved):

PR #1241, opened 2026-06-11
chore: fix deserialization in ProtoBuf of unbounded repeated booleans

richardapeters:
Deserialization of unbounded repeated booleans in Protobuf messages did not compile, due to std::vector<bool>::back() not returning a reference to a boolean

### DIFF
--- a/protobuf/echo/test/TestProtoMessageReceiver.cpp
+++ b/protobuf/echo/test/TestProtoMessageReceiver.cpp
@@ -251,6 +251,7 @@ TEST(ProtoMessageReceiverTest, parse_repeated_everything)
 
     infra::StdVectorInputStreamReader::WithStorage data;
     receiver.Feed(data);
+    EXPECT_EQ(test_messages::TestRepeatedEverything{}, receiver.message);
 }
 
 TEST(ProtoMessageReceiverTest, parse_unbounded_repeated_everything)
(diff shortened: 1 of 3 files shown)


protobuf/echo/test/TestProtoMessageReceiver.cpp
```diff
@@ -244,3 +244,21 @@ TEST(ProtoMessageReceiverTest, parse_more_nested_message)
...
     EXPECT_EQ((test_messages::TestMoreNestedMessage({ 5 }, { 10 })), receiver.message);
 }
+
+TEST(ProtoMessageReceiverTest, parse_repeated_everything)
+{
+    services::ProtoMessageReceiver<test_messages::TestRepeatedEverything> receiver;
+
+    infra::StdVectorInputStreamReader::WithStorage data;
+    receiver.Feed(data);
+    EXPECT_EQ(test_messages::TestRepeatedEverything{}, receiver.message);
```
heinwessels-philips, line 254: I would prefer if there's some data in these test messages to make it's actually parsed correctly. I know it's a lot of copy-paste values that the values print now, but this will still prevent regressions.

heinwessels-philips (approved):

protobuf/echo/test/TestProtoMessageReceiver.cpp
richardapeters: The deserialization of data happens in other tests. This test mainly tests compilability

PR #1246, opened 2026-06-18
feat: in EchoOnStreams request only as much as needed for the current packet

richardapeters:
By requesting only enough for the current packet, the Sesame stack may send the packet earlier instead of waiting for the maximum size to be available

### DIFF
--- a/protobuf/echo/EchoOnStreams.cpp
+++ b/protobuf/echo/EchoOnStreams.cpp
@@ -33,7 +33,13 @@ namespace services
 
     void EchoOnStreams::Reset()
     {
+        countingSentWriter.OnAllocatable(infra::emptyFunction);
         ResetReading();
+        really_assert(countingSentWriter.Allocatable());
+        countingSentWriter.OnAllocatable([this]()
+            {
+                StreamWriterDone();
+            });
 
         while (!sendRequesters.empty())
             sendRequesters.front().CancelRequestSend();
@@ -139,10 +145,12 @@ namespace services
 
     void EchoOnStreams::TryGrantSend()
     {
-        if (!skipNextStream && countingSentWriter.Allocatable() && sendingProxy == nullptr && !sendRequesters.empty())
+        if (!skipNextStream && sendingProxy == nullptr && !sendRequesters.empty())
         {
             sendingProxy = &sendRequesters.front();
-            sendingProxySize = sendingProxy->CurrentRequestedSize() + 2 * infra::MaxVarIntSize(std::numeric_limits<uint64_t>::max());
+            sendingProxySize = sendingProxy->CurrentRequestedSize()                              // payload
+                               + 2 * infra::MaxVarIntSize(std::numeric_limits<uint64_t>::max()); // echo framing overhead (service and method id)
+
             sendRequesters.pop_front();
             RequestSendStream(sendingProxySize);
         }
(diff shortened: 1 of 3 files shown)


protobuf/echo/EchoOnStreams.cpp
```diff
@@ -139,11 +139,12 @@ namespace services
 
     void EchoOnStreams::TryGrantSend()
     {
-        if (!skipNextStream && sendingProxy == nullptr && !sendRequesters.empty())
+        if (!skipNextStream && countingSentWriter.Allocatable() && sendingProxy == nullptr && !sendRequesters.empty())
         {
             sendingProxy = &sendRequesters.front();
+            sendingProxySize = sendingProxy->CurrentRequestedSize() + 2 * infra::MaxVarIntSize(std::numeric_limits<uint64_t>::max());
```
ccrugoPhilips:
```suggestion
            sendingProxySize = sendingProxy->CurrentRequestedSize()                              // payload
                               + 2 * infra::MaxVarIntSize(std::numeric_limits<uint64_t>::max()); // echo framing overhead
```

protobuf/echo/EchoOnStreams.hpp
```diff
@@ -49,12 +49,35 @@ namespace services
...
         void LimitedReaderDone();
 
+    private:
+        struct CountingSentWriter
+        {
+            CountingSentWriter(infra::SharedPtr<infra::StreamWriter>&& writer, std::size_t& size)
+                : writer(std::move(writer))
+                , size(size)
+            {
+                size -= this->writer->Available();
```
ccrugoPhilips, line 66:
```suggestion
                really_assert(this->size >= this->writer->Available());
                this->size -= this->writer->Available();
```

Just to make sure caller never tries to do funny things :)

protobuf/echo/EchoOnStreams.cpp
ccrugoPhilips, line 50: Maybe resetting the ` sendingProxySize` too?

protobuf/echo/EchoOnStreams.cpp
```diff
@@ -139,11 +139,12 @@ namespace services
 
     void EchoOnStreams::TryGrantSend()
     {
-        if (!skipNextStream && sendingProxy == nullptr && !sendRequesters.empty())
+        if (!skipNextStream && countingSentWriter.Allocatable() && sendingProxy == nullptr && !sendRequesters.empty())
```
ccrugoPhilips: There is no microtest covering the scenario when `countingSentWriter.Allocatable() == false` and send is not granted.

ccrugoPhilips (approved):

protobuf/echo/EchoOnStreams.cpp
richardapeters: That's not needed; in the flow, sendingProxySize is assigned before used

protobuf/echo/EchoOnStreams.hpp
richardapeters: The size could be a lot bigger than what's available in the writer. In that case, we make use of the (defined) overflow/underflow of unsigned numbers

protobuf/echo/EchoOnStreams.cpp
richardapeters: I've removed this clause because it is superfluous

PR #1253, opened 2026-06-26
feat: add DoubleBufferedSerialCommunication

richardapeters:
DoubleBufferedSerialCommunication introduces a buffer between sending data and reporting data to be sent, so that business logic can start preparing the next buffer earlier, and reduce time wasted between the end of the previous transmission and the start of the next transmission.

### DIFF
--- a/services/util/DoubleBufferedSerialCommunication.cpp
+++ b/services/util/DoubleBufferedSerialCommunication.cpp
@@ -3,7 +3,8 @@
 namespace services
 {
     DoubleBufferedSerialCommunication::DoubleBufferedSerialCommunication(infra::BoundedVector<uint8_t>& buffer, BufferedSerialCommunication& delegate)
-        : buffer(buffer)
+        : hal::BufferedSerialCommunicationObserver(delegate)
+        , buffer(buffer)
         , delegate(delegate)
     {}
 
@@ -39,9 +40,9 @@ namespace services
         if (nowActionOnCompletion != nullptr && !sending)
         {
             sending = true;
-            auto insertingChunk = infra::Head(nowSending, buffer.max_size());
-            buffer.assign(insertingChunk.begin(), insertingChunk.end());
-            nowSending = infra::DiscardHead(nowSending, insertingChunk.size());
+            auto chunk = infra::Head(nowSending, buffer.max_size());
+            buffer.assign(chunk.begin(), chunk.end());
+            nowSending = infra::DiscardHead(nowSending, chunk.size());
 
             if (nowSending.empty())
                 nowActionOnCompletion();
--- a/services/util/DoubleBufferedSerialCommunication.cpp
+++ b/services/util/DoubleBufferedSerialCommunication.cpp
@@ -1,0 +1,58 @@
+#include "services/util/DoubleBufferedSerialCommunication.hpp"
+
+namespace services
+{
+    DoubleBufferedSerialCommunication::DoubleBufferedSerialCommunication(infra::BoundedVector<uint8_t>& buffer, BufferedSerialCommunication& delegate)
+        : buffer(buffer)
+        , delegate(delegate)
+    {}
+
+    void DoubleBufferedSerialCommunication::SendData(infra::ConstByteRange data, infra::Function<void()> actionOnCompletion)
+    {
+        really_assert(actionOnCompletion != nullptr);
+        really_assert(nextActionOnCompletion == nullptr);
+
+        nextSend = data;
+        nextActionOnCompletion = actionOnCompletion;
+
+        TrySend();
+    }
+
+    infra::StreamReaderWithRewinding& DoubleBufferedSerialCommunication::Reader()
+    {
+        return delegate.Reader();
+    }
+
+    void DoubleBufferedSerialCommunication::AckReceived()
+    {
+        delegate.AckReceived();
+    }
+
+    void DoubleBufferedSerialCommunication::TrySend()
+    {
+        if (nextActionOnCompletion != nullptr && nowActionOnCompletion == nullptr)
+        {
+            nowSending = nextSend;
+            nowActionOnCompletion = std::move(nextActionOnCompletion);
+        }
+
+        if (nowActionOnCompletion != nullptr && !sending)
+        {
+            sending = true;
+            auto insertingChunk = infra::Head(nowSending, buffer.max_size());
+            buffer.assign(insertingChunk.begin(), insertingChunk.end());
+            nowSending = infra::DiscardHead(nowSending, insertingChunk.size());
+
+            if (nowSending.empty())
+                nowActionOnCompletion();
+
+            delegate.SendData(infra::MakeRange(buffer), [this]()
+                {
+                    sending = false;
+                    buffer.clear();
+
+                    TrySend();
+                });
+        }
+    }
+}
--- a/services/util/DoubleBufferedSerialCommunication.cpp
+++ b/services/util/DoubleBufferedSerialCommunication.cpp
@@ -2,7 +2,7 @@
 
 namespace services
 {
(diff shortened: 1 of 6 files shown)


services/util/DoubleBufferedSerialCommunication.hpp
```diff
@@ -0,0 +1,43 @@
+#ifndef SERVICES_DOUBLE_BUFFERED_SERIAL_COMMUNICATION_HPP
+#define SERVICES_DOUBLE_BUFFERED_SERIAL_COMMUNICATION_HPP
+
+#include "hal/interfaces/SerialCommunication.hpp"
+#include "infra/util/BoundedVector.hpp"
```
richardapeters, line 6:
```suggestion
#include "infra/util/AutoResetFunction.hpp"
#include "infra/util/BoundedVector.hpp"
```

services/util/test/TestDoubleBufferedSerialCommunication.cpp
```diff
@@ -0,0 +1,146 @@
...
+public:
+    testing::StrictMock<hal::BufferedSerialCommunicationMock> delegate;
+    services::DoubleBufferedSerialCommunication::WithStorage<4> communication{ delegate };
+    hal::BufferedSerialCommunicationObserverMock observer{ communication };
+};
+
+TEST_F(DoubleBufferedSerialCommunicationTest, read_and_ack_are_delegated)
+{
+    infra::StreamReaderWithRewindingMock reader;
+    EXPECT_CALL(delegate, Reader()).WillOnce(testing::ReturnRef(reader));
```
daantimmer, line 20: Any EXPECT_CALL that doesn't have or require or care about arguments can be written like:
```suggestion
    EXPECT_CALL(delegate, Reader).WillOnce(testing::ReturnRef(reader));
```

(omitting the empty parens)

From the docs:
> The parameter matchers... is a comma-separated list of [matchers](https://google.github.io/googletest/gmock_for_dummies.html#matchers-what-arguments-do-we-expect) that correspond to each argument of the method method_name. The expectation will apply only to calls of method_name whose arguments match all of the matchers. **If (matchers...) is omitted, the expectation behaves as if each argument’s matcher were a [wildcard matcher (_)](https://google.github.io/googletest/reference/matchers.html#wildcard)**.

It's simply a more generic way of saying 'this function must be called and we don't care how'

richardapeters: I don't like wildcard matchers :) I rather write it like this, making it instantly look like a function call, specifying that there are no parameters

services/util/test/TestDoubleBufferedSerialCommunication.cpp
```diff
@@ -0,0 +1,146 @@
...
+    EXPECT_EQ(&reader, &communication.Reader());
```
daantimmer:
```suggestion
    EXPECT_THAT(&reader, testing::Pointer(&communication.Reader()));
```

As mentioned before, use EXPECT_THAT over the old ancient EXPECT_EQ macros. You get way better error messages:

## `EXPECT_THAT(&ptr, testing::Pointer(&otherPtr));`
```
some_random_test_name.cpp:30: Failure
Value of: &foo1
Expected: is a pointer that is equal to 0x7ffe594501d7
  Actual: 0x7ffe594501d6 (of type Foo*), which is a pointer that 0x7ffe594501d6 (of type Foo const*)
```

## `EXPECT_THAT(&ptr, testing::Eq(&otherPtr));`
```
some_random_test_name.cpp:31: Failure
Value of: &foo1
Expected: is equal to 0x7ffe594501d7
  Actual: 0x7ffe594501d6 (of type Foo*)
```

## `EXPECT_EQ(&ptr, &otherPtr);`
```
some_random_test_name.cpp:32: Failure
Expected equality of these values:
  &foo1
    Which is: 0x7ffe594501d6
  &foo2
    Which is: 0x7ffe594501d7
```

daantimmer: same applies for all the other EXPECT_EQ and or other derivates

services/util/test/TestDoubleBufferedSerialCommunication.cpp
```diff
@@ -0,0 +1,146 @@
...
+    EXPECT_CALL(delegate, SendData(infra::ContentsEqual(data), testing::_)).WillOnce(testing::SaveArg<1>(&onSent));
+    communication.SendData(data, [&]()
+        {
+            onDone.callback();
+        });
+
+    // Verify that the onDone callback has happened before the delegate's onDone is reported
+    testing::Mock::VerifyAndClearExpectations(&onDone);
+    onSent();
+}
```
daantimmer, line 50: This should use checkpoints instead:

```suggestion
{
    static const std::array<uint8_t, 4> data{ 1, 2, 3, 4 };
    testing::MockFunction<void(std::string_view checkPointName)> check;
    testing::StrictMock<infra::MockCallback<void()>> onDone;
    infra::Function<void()> onSent;
    
    {
        testing::InSequence s;

        EXPECT_CALL(onDone, callback());
        EXPECT_CALL(check, Call("1"));
        EXPECT_CALL(delegate, SendData(infra::ContentsEqual(data), testing::_)).WillOnce(testing::SaveArg<1>(&onSent));
        EXPECT_CALL(check, Call("2"));
    }

    
    communication.SendData(data, [&]()
        {
            onDone.callback();
        });
    check("1");
    
    onSent();
    check("2");
}
```

I am not sure about the placement of the check::Call expectation, whether it should be before or after the EXPECT_CALL(delegate, SendData). You'll be more familiar with that.

services/util/test/TestDoubleBufferedSerialCommunication.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+TEST_F(DoubleBufferedSerialCommunicationTest, send_data_is_delegated_and_reported_done_immediately_when_it_fits_in_the_buffer)
+{
+    testing::InSequence s;
+
+    static const std::array<uint8_t, 4> data{ 1, 2, 3, 4 };
+    testing::StrictMock<infra::MockCallback<void()>> onDone;
+    infra::Function<void()> onSent;
+    EXPECT_CALL(onDone, callback());
+    EXPECT_CALL(delegate, SendData(infra::ContentsEqual(data), testing::_)).WillOnce(testing::SaveArg<1>(&onSent));
```
daantimmer:
```suggestion
    EXPECT_CALL(delegate, SendData(testing::ElementsAreArray(data), testing::_)).WillOnce(testing::SaveArg<1>(&onSent));
```

services/util/DoubleBufferedSerialCommunication.cpp
```diff
@@ -0,0 +1,65 @@
...
+        if (nextActionOnCompletion != nullptr && nowActionOnCompletion == nullptr)
+        {
+            nowSending = nextSend;
+            nowActionOnCompletion = std::move(nextActionOnCompletion);
+        }
+
+        if (nowActionOnCompletion != nullptr && !sending)
+        {
+            sending = true;
+            auto chunk = infra::Head(nowSending, buffer.max_size());
```
daantimmer, line 49:
```suggestion
            const auto chunk = infra::Head(nowSending, buffer.max_size());
```

services/util/DoubleBufferedSerialCommunication.hpp
daantimmer, line 1: I am finding the lack of consistency, interesting.
- I thought we agreed to not use multiple same-visibility sections (multiple `public:` or `private:` sections?
- If we didn't there is still an inconsistency in how the public and private parts  are written. Where the private separates overriden methods from non-overriden methods. Yet this is not done for the public section.

richardapeters: Multiple access blocks are definitely encouraged! They work great to create blocks. I've added a few to make it more consistent, tough.

daantimmer (approved):

PR #1254, opened 2026-06-26
fix: windows network race condition: defer FD_CLOSE until data drained

brianvanderbeek-philips:
WinConnection: defer FD_CLOSE until data drained


services/network/HttpClientBasic.cpp
```diff
@@ -113,10 +113,20 @@ namespace services
...
+                        Expire();
+                    });
+            }
+            else
+            {
+                sharedAccess.SetAction([this]()
+                    {
+                        ReportError(true);
+                    });
+            }
```
richardapeters: HttpClientBasic::Attached() already set this action to Expire(), so this does not need to be overwritten, right?

```suggestion
            if (!receivedStatus || contentError)
            {
                sharedAccess.SetAction([this]()
                    {
                        ReportError(true);
                    });
            }
```

brianvanderbeek-philips: Yes, I think you are correct. I will make that change.

Does this mean you also agree with the general concept of not setting the action to ReportError(true) in this case? If so, then I will update the unit tests accordingly.

services/network_instantiations/ConnectionWin.cpp
```diff
@@ -250,6 +198,42 @@ namespace services
...
     }
 
+    void ConnectionWin::ScheduleDataReceivedIfNeeded()
+    {
+        if (dataReceivedScheduled)
+            return;
+
+        dataReceivedScheduled = true;
+        infra::EventDispatcherWithWeakPtr::Instance().Schedule([](const infra::SharedPtr<ConnectionWin>& object)
+            {
```
richardapeters, line 276: I'd recommend extracting this as a function, so that "object->" is not necessary.

services/network_instantiations/ConnectionWin.cpp
```diff
@@ -250,6 +198,42 @@ namespace services
...
+                if (object->IsAttached())
+                {
+                    object->Observer().DataReceived();
+
+                    auto bufferedAfterCallback = object->receiveBuffer.size();
+                    if (object->closePending && bufferedAfterCallback != 0 && bufferedAfterCallback < bufferedBeforeCallback)
+                    {
+                        object->ScheduleDataReceivedIfNeeded();
+                        return;
+                    }
```
richardapeters: Why is this necessary? I would expect that a client had the opportunity to read the received data. If it didn't complete it, it should be safe to close the connection.

services/network_instantiations/ConnectionWin.cpp
```diff
@@ -191,6 +198,42 @@ namespace services
...
+            SharedFromThis());
+    }
+
+    void ConnectionWin::FinalizeCloseIfReady()
+    {
+        if (closePending && !dataReceivedScheduled)
+        {
+            CloseAndDestroy();
+        }
+    }
```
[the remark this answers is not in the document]
brianvanderbeek-philips: If the observer does not consume any bytes from the buffer then we should close. In this case the buffer would not be empty. I don't think there should be a forced buffer drain before close.

services/network_instantiations/ConnectionWin.cpp
brianvanderbeek-philips: Ok, will do

services/network_instantiations/ConnectionWin.cpp
brianvanderbeek-philips: When initially calling DataReceived once and then handling the close, not all data was consumed and the introspect still failed. When doing the close only after DataReceived() is no longer processing the data solves that. Here is a trace log snippet:

ConnectionMbedTls::DataReceived mbedtls_ssl_read result=1024 buffered-before-resize=0
ConnectionMbedTls::DataReceived decrypted bytes=1024 buffered-now=1024
ConnectionMbedTls::DataReceived scheduling observer callback buffered=1024
ConnectionWin::ProcessDataReceived end buffered=24 progress=1               <--- remaining 24B
ConnectionWin::ProcessDataReceived scheduling follow-up DataReceived 
ConnectionMbedTls::DataReceived invoking observer DataReceived buffered=1024
HttpClientImpl::DataReceived bodyReaderActive=0 responseParserActive=1
HttpClientImpl::HandleData headerParsingDone=0
HttpClientImplWithRedirection::StatusAvailable HTTP/1.1 200 OK
HttpClientImpl::StatusAvailable code=200 statusLine=HTTP/1.1 200 OK
HttpClientImpl::HandleData parser step done

services/network/HttpClientBasic.cpp
```diff
@@ -113,10 +113,14 @@ namespace services
         {
             state = State::closing;
             timeoutTimer.Cancel();
-            sharedAccess.SetAction([this]()
-                {
-                    ReportError(true);
-                });
+
+            if (!receivedStatus || contentError)
```
richardapeters: Thinking about this again, I thing this is the incorrect place to fix the bug. The question is not "if we are detaching, do we report an error once HttpClient drops us", but "if we are detaching, we ReportError if we are droppen, _unless_ we find out a little later that the body has actually been completely and correctly parsed".

So instead of adding the if-statement on this line, I think the bug is better fixed by restoring
        sharedAccess.SetAction([this]()
            {
                Expire();
            });
in BodyComplete()

services/network_instantiations/ConnectionWin.cpp
```diff
@@ -250,6 +198,47 @@ namespace services
...
+    void ConnectionWin::ProcessDataReceived()
+    {
+        auto bufferedBeforeCallback = receiveBuffer.size();
+
+        dataReceivedScheduled = false;
+        if (IsAttached())
+        {
+            Observer().DataReceived();
+
+            auto bufferedAfterCallback = receiveBuffer.size();
```
richardapeters: Observer().DataReceived() signals the observer that it can obtain a shared pointer to a reader. The observer can then take its time to use that reader. The next few lines here should therefore not be executed immediately, but upon either the reader becoming allocatable (in the constructor of ConnectionWin the streamReader is initialized with an onAllocatable), or when during the call to Observer().DataReceived() the reader has not yet been obtained.

services/network_instantiations/ConnectionWin.cpp
```diff
@@ -250,6 +198,47 @@ namespace services
...
+
+        FinalizeCloseIfReady();
+    }
+
+    void ConnectionWin::FinalizeCloseIfReady()
+    {
+        if (closePending && !dataReceivedScheduled)
+        {
+            CloseAndDestroy();
+        }
```
richardapeters:
```suggestion
            CloseAndDestroy();
```

services/network/HttpClientBasic.cpp
richardapeters: Thinking about this yet again, why are we receiving a Detaching() while the data is not yet completely in? Shouldn't this be solved in ConnectionMbedTls, by having that class not detach before pending data is made available?

richardapeters (approved):

PR #1257, opened 2026-06-29
chore: ensure MemoryRange constructed with dangling pointer does not compile

heinwessels-philips:
This makes the `MemoryRange` more safe by ensuring the memory don't easily contain dangling references. This is case that the compiler clearly should not allow. I stumbled over this twice recently while copy-pasting code from various sources.

```C++
auto danglingPointer = infra::MakeRange({ 5 });
// At this point `{ 5 }` has been destructed, and the `MemoryRange` is now pointing to garbage.
```

Edit: I've verified that all our upstream code still compiles with this change.


infra/util/test/TestMemoryRange.cpp
```diff
@@ -312,6 +313,15 @@ TEST(MemoryRangeTest, TestMakeRangeFromContainer)
     EXPECT_EQ(infra::ByteRange(&container.front(), &container.front() + 3), infra::MakeConstRange(static_cast<const infra::BoundedVector<uint8_t>&>(container)));
 }
 
+TEST(MemoryRangeTest, TestMakeRangeFromInitializerList)
+{
+    std::initializer_list<uint8_t> list{ 2, 3, 4 };
+    EXPECT_THAT(infra::MakeRange(list), testing::ElementsAreArray({ 2, 3, 4 }));
```
daantimmer:
```suggestion
    EXPECT_THAT(infra::MakeRange(list), testing::ElementsAre(2, 3, 4));
```

daantimmer (approved):

PR #1266, opened 2026-07-09
chore: really_assert if 'SynchronousFlashRegion' is defined correctly

heinwessels-philips:
Previously this was a regular `assert`, which was missed for a month already because nobody uses debug builds.


richardapeters (approved):

fabiangottstein:
@heinwessels-philips This codebase uses a lot `assert` instead of `really_assert`. Would you replace all to align with this PR?

heinwessels-philips:
> heinwessels (@heinwessels-philips) This codebase uses a lot `assert` instead of `really_assert`. Would you replace all to align with this PR?

I agree that it should be changed, but a big change like this would first need to be discussed with internally first to weigh the pros and cons.

PR #1267, opened 2026-07-10
feat: enable lwip checksum in software as option

andjordan:
feat: enable lwip checksum in software as option
Some network stacks/hardware does not compute the checksum, which requires lwip do it by software

fix: communication stall after TLS session ticket. 
mbedTLS can consume encrypted bytes from the receive buffer yet still return WANT_READ because it doesn't have a complete TLS record to yield application data yet. Previously the loop would unconditionally break and wait for the next DataReceived() notification. But since the underlying data was already consumed/acked in that pass, no new notification necessarily arrives — so buffered encrypted data could sit unprocessed and the connection would stall/deadlock.

By tracking whether the BIO provided data and looping again when it did, the code keeps draining available encrypted data instead of giving up prematurely, preventing the stall.


lwip/lwip_config/CMakeLists.txt
```diff
@@ -9,3 +9,10 @@ target_include_directories(lwip.lwip_conf INTERFACE
 target_sources(lwip.lwip_conf PRIVATE
     sys_now.cpp
 )
+
+if (EMIL_LWIP_CHECKSUM_IN_SOFTWARE)
+    message(STATUS "Enabling lwIP checksum calculation in software")
```
andjordan:
```suggestion
```

richardapeters (approved):

PR #1272, opened 2026-07-14
feat: move hal/interfaces/BleDtm -> services/ble/Dtm

daantimmer:
BleDTm is not a HAL interface, but a service interface. Therefor its moved to services/ble and renamed from BleDtm to just Dtm

### DIFF
--- a/services/ble/Dtm.hpp
+++ b/services/ble/Dtm.hpp
@@ -9,10 +9,13 @@ namespace services
     class Dtm
     {
     public:
+    protected:
         Dtm() = default;
+        ~Dtm() = default;
+
+    public:
         Dtm(const Dtm& other) = delete;
         Dtm& operator=(const Dtm& other) = delete;
-        ~Dtm() = default;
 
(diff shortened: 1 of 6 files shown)


services/ble/Dtm.hpp
```diff
@@ -1,18 +1,18 @@
...
     {
     public:
-        BleDtm() = default;
-        BleDtm(const BleDtm& other) = delete;
-        BleDtm& operator=(const BleDtm& other) = delete;
-        ~BleDtm() = default;
+        Dtm() = default;
+        Dtm(const Dtm& other) = delete;
+        Dtm& operator=(const Dtm& other) = delete;
+        ~Dtm() = default;
```
fabiangottstein:
```suggestion
    protected:
        Dtm() = default;
        ~Dtm() = default;

    public:
        Dtm(const Dtm& other) = delete;
        Dtm& operator=(const Dtm& other) = delete;
```

EkelmansPh (approved):

PR #1291, opened 2026-08-03
docs: update GapPeripheral and GapPeripheralResponse

MiquelJayson-Philips:
Clearer documentation regarding the GapPeripheral Service.


services/ble/Gap.proto
```diff
@@ -238,62 +234,70 @@ message SecurityPreference
...
 
-    // Allowed in link layer states: advertising, connected
+    // Allowed states: standby, advertising, connected
     // Results in GapPeripheralResponse.CurrentState with state standby
     rpc Standby(Nothing) returns (Nothing) { option (method_id) = 3; }
 
-    // Allowed in link layer states: standby
-    // Formatting of advertisement data is as per Bluetooth Core Specification.
+    // Allowed states: standby
+    // Formatting of advertisement data is as per Bluetooth Core Specification v6.3, Vol 3, Part C, Section 11
```
heinwessels-philips: The location in the specification should probably be added at the `message AdvertisementData` instead

services/ble/Gap.proto
```diff
@@ -238,62 +234,70 @@ message SecurityPreference
...
     // Maximum advertisement data length for this method is 28 bytes. Server reserves an additional 3 bytes for flags data with the flag bits 'LE General Discoverable Mode' and 'BR/EDR Not Supported' enabled.
     // Server will not check validity of the advertisement data and only maximum length of each field will be passed to BLE stack
+    // If set, the advertisement data remains in effect until it is overwritten or the device is rebooted.
     // If not set by this method, server will have as advertisement data only flags field.
     rpc SetAdvertisementData(AdvertisementData) returns (Nothing) { option (method_id) = 4; }
 
-    // Allowed in link layer states: standby
-    // Formatting of scan response data is as per Bluetooth Core Specification.
+    // Allowed states: standby
+    // Formatting of scan response data is as per Bluetooth Core Specification v6.3, Vol 3, Part C, Section 11
```
heinwessels-philips: Same here, specification should be at the `message ScanResponseData`

services/ble/Gap.proto
```diff
@@ -238,62 +234,70 @@ message SecurityPreference
...
     // If not set by this method, server will have as scan response data only service uuids.
     rpc SetScanResponseData(ScanResponseData) returns (Nothing) { option (method_id) = 5; }
     
-    // Allowed in link layer states: advertising
+    // Allowed states: standby, advertising, connected
     // Results in GapPeripheralResponse.AdvertisingAddress
-    // Type of the advertising address is statically configured. It is the same with Identity Address when privacy is 
-    // not used and different otherwise.
+    // When privacy is disabled, it matches the identity address; when privacy is enabled, it returns the resolvable private address generated by the BLE stack
+    // If GapPeripheral.Advertise was never called and the device is in standby, returns the identity address.
```
heinwessels-philips:
```suggestion
    // If GapPeripheral.Advertise was never called and the device is in standby, always returns the identity address.
```

services/ble/Gap.proto
```diff
@@ -238,62 +234,70 @@ message SecurityPreference
...
+    // Once advertised it will return the latest advertising address
     rpc GetAdvertisingAddress(Nothing) returns (Nothing) { option (method_id) = 6; }
 
+    // Allowed states: standby, advertising, connected
     // Results in GapPeripheralResponse.IdentityAddress
-    // Type of the identity address is statically configured
+    // Type of the identity address(public/random) is statically configured
```
heinwessels-philips: This sentence is a bit hard to read.
```suggestion
    // Type of the identity address (i.e. public or random) is statically configured
```
Also, what does `statically configured` mean? If I'm a customer and I this I might struggle to know where to check if this function would return the public or random address.

services/ble/Gap.proto
```diff
@@ -341,10 +344,16 @@ service GapPeripheralResponse
 
     rpc PairingResult(PairingStatus) returns (Nothing) { option (method_id) = 5; }
 
+    // NumberOfBondsChanged is triggered when the bond list is updated after bonding or bond removal.
     rpc NumberOfBondsChanged(UInt32Value) returns(Nothing) { option(method_id) = 6; }
-
+    
     rpc BondInformation(BondsStatus) returns(Nothing) { option(method_id) = 7; }
 
+    // Sent during an active pairing/bonding procedure when user interaction is required to complete pairing.
```
heinwessels-philips, line 363: Maybe we can also add in which states they might expect this call? Same for other responses.

```suggestion
    // Expected states: connected
    // Sent during an active pairing/bonding procedure when user interaction is required to complete pairing.
```

heinwessels-philips (changes requested):

services/ble/Gap.proto
MiquelJayson-Philips: I like that idea, will apply it!

services/ble/Gap.proto
MiquelJayson-Philips: I rephrased it to "// The address type (public or random) is determined by the HAL implementation and is not configurable via the service."

services/ble/Gap.proto
```diff
@@ -303,48 +307,64 @@ service GapPeripheral
...
+    // Expected states: standby, advertising, connected
+    // Sent in response to GapPeripheral.GetBondInformation.
     rpc BondInformation(BondsStatus) returns(Nothing) { option(method_id) = 7; }
 
+    // Expected states: connected
+    // Sent during an active pairing/bonding procedure when user interaction is required to complete pairing.
+    // This will only be triggered when the negotiated pairing method requires it, which is determined by the
+    // I/O capabilities configured via GapPeripheral.SetSecurityPreferences
+    // Not all pairing methods require user interaction (e.g. Just Works will never trigger this).
+    // Must be responded to with GapPeripheral.NumericComparisonConfirm when AuthenticationRequest.isNumericComparison is true.
```
heinwessels-philips, line 367: This must be added to the `GapCentralResponse::AuthenticationRequired` if you're removing it from the top.

services/ble/Gap.proto
```diff
@@ -303,48 +307,64 @@ service GapPeripheral
...
+    // Expected states: standby, advertising, connected
+    // Sent in response to GapPeripheral.GetIdentityAddress.
     rpc IdentityAddress(AddressWithType) returns(Nothing) { option(method_id) = 4; }
 
+    // Expected states: connected
+    // Sent when a pairing/bonding procedure completes or fails.
     rpc PairingResult(PairingStatus) returns (Nothing) { option (method_id) = 5; }
 
+    // Expected states: standby, connected
+    // NumberOfBondsChanged is triggered when the bond list is updated after bonding or bond removal.
```
heinwessels-philips: This response is currently also sent in some cases when bonding occurs, even if the bond was pre-existing. Meaning nothing was `updated`. Or does `updated` also mean that? Might be that I'm misreading `GapTiCC35` though

heinwessels-philips (approved):

services/ble/Gap.proto
MiquelJayson-Philips: Will be added to GapCentralResponse service in gap.proto

services/ble/Gap.proto
MiquelJayson-Philips: Rephrased it to "// Triggered when a new bond is created, an existing bond is removed, or re-bonding occurs with an already bonded peer." this should be more explicit. Instead of just updated.

PR #1292, opened 2026-08-04
fix: http path and headers larger than available stream size

andjordan:
This PR fixes HTTP request transmission when the formatted request line + headers exceed the maximum send stream size provided by the connection, ensuring headers (and long request targets) are emitted correctly across multiple send-stream allocations.

**Changes:**
- Cap initial `RequestSendStream()` sizes to `MaxSendStreamSize()` in `HttpClientImpl`, avoiding oversize send-stream requests.
- Make `HttpRequestFormatter` able to write/consume headers incrementally via a running `headerPosition`, enabling header splitting across multiple send streams.
- Add tests that cover headers exceeding the send stream size (including custom headers).

### DIFF
--- a/services/network/Http.cpp
+++ b/services/network/Http.cpp
@@ -39,6 +39,29 @@ namespace services
 
             return schemeEnd;
         }
+
+        void WriteFragment(infra::TextOutputStream& stream, infra::BoundedConstString fragment, std::size_t& skip, std::size_t& written)
+        {
+            if (skip >= fragment.size())
+                skip -= fragment.size();
+            else
+            {
+                fragment = fragment.substr(skip);
+                skip = 0;
+
+                auto amount = std::min(stream.Available(), fragment.size());
+                stream << fragment.substr(0, amount);
+                written += amount;
+            }
+        }
+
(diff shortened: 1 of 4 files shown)


services/network/Http.cpp
```diff
@@ -160,6 +144,60 @@ namespace services
...
+    std::size_t HttpRequestFormatter::HeaderBlockSize() const
+    {
+        return HttpVerbToString(verb).size() + requestTarget.size() + httpVersion.size() + HeadersSize() + (2 * crlf.size()) + (2 * sp.size());
+    }
+
+    std::size_t HttpRequestFormatter::WriteHeader(infra::TextOutputStream& stream, std::size_t skip) const
+    {
+        std::size_t written = 0;
+
+        auto writeFragment = [&stream, &skip, &written](infra::BoundedConstString fragment)
```
BarisTanyeri: Why not a private function for this, instead of a lambda ?

BarisTanyeri (approved):

services/network/Http.cpp
andjordan: extracted for readability (compiler end result is nearly the same)

PR #1295, opened 2026-08-05
feat(ble): add BLE bond name storage interface

heinwessels-philips:
This adds an interface to store the name of an device as part of a BLE bond.


danielschenk (changes requested):
agree on the AI comments

danielschenk (approved):

PR #1298, opened 2026-08-06
feat: update StdStringInputStream to use std::string_view

daantimmer:
`StdStringInputStream` only reads from string-like object. Therefor `std::string_view` is more lean and also accepts literal strings without creating an 'expensive' `std::string` object.


danielschenk (changes requested):
Agree on the AI comment

danielschenk (approved):

PR #1302, opened 2026-08-10
feat: log error when aborting on uninitialized function call

heinwessels-philips:
Previously when a null `infra::Function` was executed there would be no abort message, just a simple `std::abort`. This made it hard to determine what went wrong on glance, and you had to dig through the stack trace and look at entries in the `Function.cpp` which is hard to read.

This changes it so that instead of calling abort it will call `LOG_AND_ABORT(...)` which will log an helpful message first.


MiquelJayson-Philips (approved):
LGTM

PR #1304, opened 2026-08-11
feat!: refresh GattClient proto interface and solve race conditions

danielschenk:
Updates GattClient(Response) with improved documentation to remove ambiguities, and improves the interface to be able to give some more guarantees and solve race conditions.

# Proto Ducky findings used for this task
image

### DIFF
--- a/services/ble/Gatt.proto
+++ b/services/ble/Gatt.proto
@@ -110,7 +110,7 @@ service GattClient
     // Responses: zero or more GattClientResponse.ServiceDiscovered, followed by GattClientResponse.ServiceDiscoveryComplete.
     rpc StartServiceDiscovery(Nothing) returns (Nothing) { option (method_id) = 2; }
     // Discovers all characteristics within a single service's handle range.
-    // Must be called once per discovered service, sequentially, after GattClientResponse.ServiceDiscoveryComplete.
+    // Must be called once per discovered service, after GattClientResponse.ServiceDiscoveryComplete.
     // Responses: zero or more GattClientResponse.CharacteristicDiscovered, followed by GattClientResponse.CharacteristicDiscoveryComplete.
     rpc StartCharacteristicDiscovery(HandleRange) returns (Nothing) { option (method_id) = 3; }
     // Discovers all descriptors within the given handle range of a characteristic.
--- a/services/ble/Gatt.proto
+++ b/services/ble/Gatt.proto
@@ -62,7 +62,8 @@ message Characteristic
     uint32 properties = 4;
 }
 
-message CharacteristicDescriptor
+//'Descriptor' keyword is not allowed by Echo code generator
+message Descriptr
 {
     bytes uuid = 1 [(bytes_size) = 16];
     Handle handle = 2;
@@ -132,8 +133,8 @@ service GattClient
     //   The handle range should be the service's start and end handles.
     rpc StartCharacteristicDiscovery(HandleRange) returns (Nothing) { option (method_id) = 3; }
 
-    // Expected response: GattClientResponse.CharacteristicDescriptorDiscoveryComplete
-    // Additional responses: zero or more GattClientResponse.CharacteristicDescriptorDiscovered
+    // Expected response: GattClientResponse.DescriptorDiscoveryComplete
+    // Additional responses: zero or more GattClientResponse.DescriptorDiscovered
     // Description: Discovers all descriptors within the given handle range of a characteristic.
     //   The handle range spans from the characteristic's value handle to the next characteristic's declaration handle - 1
     //   (or the service's end handle for the last characteristic in a service).
@@ -154,10 +155,10 @@ service GattClient
 
     // Expected response: GattClientResponse.OperationComplete
     // Additional responses: GattClientResponse.ReadDescriptorResponse
-    rpc ReadCharacteristicDescriptor(Handle) returns (Nothing) { option (method_id) = 8; }
+    rpc ReadDescriptor(Handle) returns (Nothing) { option (method_id) = 8; }
 
     // Expected response: GattClientResponse.OperationComplete
-    rpc WriteCharacteristicDescriptor(AttributeData) returns (Nothing) { option (method_id) = 9; }
+    rpc WriteDescriptor(AttributeData) returns (Nothing) { option (method_id) = 9; }
 
     // Description: Must be called for every GattClientResponse.IndicationReceived to acknowledge the indication.
     rpc IndicationDone(Nothing) returns (Nothing) { option (method_id) = 10; }
@@ -184,18 +185,18 @@ service GattClientResponse
     rpc CharacteristicDiscoveryComplete(Nothing) returns (Nothing) { option (method_id) = 5; }
 
     // Triggered by: GattClient.StartDescriptorDiscovery (for each discovered descriptor)
-    rpc CharacteristicDescriptorDiscovered(CharacteristicDescriptor) returns (Nothing) { option (method_id) = 6; }
+    rpc DescriptorDiscovered(Descriptr) returns (Nothing) { option (method_id) = 6; }
 
     // Triggered by: GattClient.StartDescriptorDiscovery (on completion)
-    rpc CharacteristicDescriptorDiscoveryComplete(Nothing) returns (Nothing) { option (method_id) = 7; }
+    rpc DescriptorDiscoveryComplete(Nothing) returns (Nothing) { option (method_id) = 7; }
 
     // Triggered by: GattClient.ReadCharacteristic
     rpc ReadCharacteristicResponse(ReadResult) returns (Nothing) { option (method_id) = 8; }
 
-    // Triggered by: GattClient.ReadCharacteristicDescriptor
+    // Triggered by: GattClient.ReadDescriptor
     rpc ReadDescriptorResponse(ReadResult) returns (Nothing) { option (method_id) = 9; }
 
-    // Triggered by: GattClient.ReadCharacteristic, GattClient.WriteCharacteristic, GattClient.WriteCharacteristicWithoutResponse, GattClient.ReadCharacteristicDescriptor, GattClient.WriteCharacteristicDescriptor
+    // Triggered by: GattClient.ReadCharacteristic, GattClient.WriteCharacteristic, GattClient.WriteCharacteristicWithoutResponse, GattClient.ReadDescriptor, GattClient.WriteDescriptor
     // Description: Reports the outcome of a read or write operation.
     rpc OperationComplete(OperationResult) returns (Nothing) { option (method_id) = 10; }
 
--- a/services/ble/Gatt.proto
+++ b/services/ble/Gatt.proto
@@ -143,22 +143,26 @@ service GattClient
 
     // Expected response: GattClientResponse.OperationComplete
     // Additional responses: GattClientResponse.ReadCharacteristicResponse
-    // Description: Read/write operations are sequential; wait for GattClientResponse.OperationComplete before issuing the next.
+    // Description: Reads the value of a characteristic from the GATT server.
     rpc ReadCharacteristic(Handle) returns (Nothing) { option (method_id) = 5; }
 
     // Expected response: GattClientResponse.OperationComplete
+    // Description: Writes the value of a characteristic to the GATT server.
     rpc WriteCharacteristic(AttributeData) returns (Nothing) { option (method_id) = 6; }
 
     // Expected response: GattClientResponse.OperationComplete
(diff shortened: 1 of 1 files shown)


services/ble/Gatt.proto
```diff
@@ -150,10 +150,9 @@ service GattClientResponse
     rpc ReadCharacteristicResponse(ReadResult) returns (Nothing) { option (method_id) = 8; }
     rpc ReadDescriptorResponse(ReadResult) returns (Nothing) { option (method_id) = 9; }
 
-    // Called after completion of operations with response: ReadCharacteristic, WriteCharacteristic, ReadDescriptor, WriteDescriptor
+    // Called after completion of characteristic operations
```
oguzcanoguz: Please make it clear which operations are the 'characteristic' operations. 
It is not obvious that WriteDescriptor is one.

services/ble/Gatt.proto
```diff
@@ -150,10 +150,9 @@ service GattClientResponse
...
     rpc OperationComplete(OperationResult) returns (Nothing) { option (method_id) = 10; }
 
     rpc IndicationReceived(AttributeData) returns (Nothing) { option (method_id) = 11; }
     rpc NotificationReceived(AttributeData) returns (Nothing) { option (method_id) = 12; }
```
oguzcanoguz, line 225: Do we have a concern here similar to the flow control of WriteWithoutResponse?
What if GattClient receives a bunch of notifications which it stores and cannot pass to the client?

danielschenk: It seems only one outstanding indication is allowed per ATT bearer. I assume we only have one such bearer per connection. So I think we are safe

danielschenk: Just realizing you meant notifications, sorry. Those don't have that flow control built in of course. I think it will be hard to do something about that. If the client can't receive them fast enough, everything will stall and eventually that can become a problem for the peripheral, yes.

oguzcanoguz: It won't be a problem for the peripheral but for central(us). Since our ble stack will push them to us as fast as possible but we might not be able to send all of them them via NotificationReceived. Peripheral wont be aware of this if there is no other channel to let it know.

However, thinking now, this is not different from running an application on cc35 itself without a proto rpc client. There also the fw might not be able process the received notifications fast enough, in which case ble stack will start dropping them.

oguzcanoguz:
Regarding the points in the agent output:

- 5.3 We should document that misuse will result in an assert in the very same way we do in Gap.proto
- 5.7 We can state that the max number of items are configuration specific
- 5.8 and 5.9 are defined by ble spec, no need to repeat here
- 5.1 and 5.6. I am curious about this. MTU exchange might conflict with other operations. We need to check and document. 
- We need to document that GattClient service is only available after connection is established. 
- "// No responses are sent after a connection loss is reported by the GAP service" this should be a guarantee we are giving.

services/ble/Gatt.proto
danielschenk: From Elia:
- works only without the +1 (or -1)
- why limit to this range and not all characteristics of service?

services/ble/Gatt.proto
```diff
@@ -109,7 +110,7 @@ service GattClient
     // Responses: zero or more GattClientResponse.ServiceDiscovered, followed by GattClientResponse.ServiceDiscoveryComplete.
     rpc StartServiceDiscovery(Nothing) returns (Nothing) { option (method_id) = 2; }
```
danielschenk, line 130: From Elia:
- first characteristic is the service declaration

danielschenk, line 130: From Brian: add an example? Oguzcan: sequence diagram

services/ble/Gatt.proto
```diff
@@ -119,7 +120,6 @@ service GattClient
     rpc StartDescriptorDiscovery(HandleRange) returns (Nothing) { option (method_id) = 4; }
 
     // Read/write operations are sequential; wait for GattClientResponse.OperationComplete before issuing the next.
-    // WriteCharacteristicWithoutResponse is fire-and-forget and does not trigger GattClientResponse.OperationComplete.
     rpc ReadCharacteristic(Handle) returns (Nothing) { option (method_id) = 5; }
     rpc WriteCharacteristic(AttributeData) returns (Nothing) { option (method_id) = 6; }
     rpc WriteCharacteristicWithoutResponse(AttributeData) returns (Nothing) { option (method_id) = 7; }
```
danielschenk, line 159: From Elia: document delivery guarantees to peripheral?

services/ble/Gatt.proto
danielschenk: From Elia: confusing
- Daniel: let's see how we can improve this (can 2 be called `declarationHandle`?)
- Brian: check descriptor terminology as well (writing CCCD uses `handle`)

danielschenk, line 168: Brian: name `*CharacteristicDescriptor`?

danielschenk: - Characteristic: renamed handle to declarationHandle for preciseness
- AttributeData::handle: leaving, `handle` is the best name already. For characteristic operations it will be the value handle. For descriptor operations it will be the attribute handle (single attribute only). This can be best generalized to attribute handle (which it already gets, because of the enclosing scope)

services/ble/Gatt.proto
```diff
@@ -94,66 +95,129 @@ enum CccdSetting
...
-    // Discovers all descriptors within the given handle range of a characteristic.
-    // The handle range spans from the characteristic's value handle + 1 to the next characteristic's handle - 1
-    // (or the service's end handle for the last characteristic in a service).
-    // Responses: zero or more GattClientResponse.DescriptorDiscovered, followed by GattClientResponse.DescriptorDiscoveryComplete.
+
+    // Expected response: GattClientResponse.DescriptorDiscoveryComplete
+    // Additional responses: zero or more GattClientResponse.DescriptorDiscovered
+    // Description: Discovers all descriptors within the given handle range of a characteristic.
+    //   The handle range spans from the characteristic's value handle to the next characteristic's declaration handle - 1
+    //   (or the service's end handle for the last characteristic in a service).
```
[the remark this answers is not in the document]
danielschenk: GATT spec clearly states that the range for this operation should start on the value handle

services/ble/Gatt.proto
```diff
@@ -94,66 +95,135 @@ enum CccdSetting
...
+//  At most one operation may be in progress at any time, with the exception of WriteCharacteristicWithoutResponse,
+//  which may overlap with any other in-progress operation except itself.
+//  Non-conformance will result in an assertion.
+// Discovery constraints:
+//  Discovery must follow the sequence: Service -> Characteristic -> Descriptor.
+//  Each step must complete (indicated by the corresponding *Complete response) before initiating the next.
+//  The maximum number of items returned depends on the configuration.
+// Connection loss:
+//  Any pending discovery or read/write operation is implicitly cancelled.
+//  For any pending operation, a corresponding response with an error status is guaranteed.
```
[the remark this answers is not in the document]
danielschenk: handled the MTU case now too

services/ble/Gatt.proto
```diff
@@ -94,66 +95,136 @@ enum CccdSetting
...
+    // Description: Reports whether a write without response command was successfully queued by the GATT client for transmission to the GATT server.
+    //  Success indicates the request was processed by the GATT client, but does not guarantee success on the GATT server side.
+    rpc WriteCharacteristicWithoutResponseComplete(OperationResult) returns (Nothing) { option (method_id) = 10; }
 
-    rpc IndicationReceived(AttributeData) returns (Nothing) { option (method_id) = 11; }
-    rpc NotificationReceived(AttributeData) returns (Nothing) { option (method_id) = 12; }
-}
+    // Triggered by: GattClient.ReadCharacteristic, GattClient.WriteCharacteristic, GattClient.ReadDescriptor, GattClient.WriteDescriptor
+    // Description: Reports the outcome of a read or write operation.
+    rpc OperationComplete(OperationResult) returns (Nothing) { option (method_id) = 11; }
```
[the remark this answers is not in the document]
danielschenk: message types are already changing, therefore accepting that everything breaks and taking opportunity to reorder

services/ble/Gatt.proto
```diff
@@ -94,66 +95,136 @@ enum CccdSetting
...
     rpc MtuSizeExchangeRequest(Nothing) returns (Nothing) { option (method_id) = 1; }
 
-    // Discovery must follow the sequence: Service -> Characteristic -> Descriptor.
-    // Each step must complete (indicated by the corresponding *Complete response) before initiating the next.
-    //
-    // Discovers all services on the remote device.
-    // Responses: zero or more GattClientResponse.ServiceDiscovered, followed by GattClientResponse.ServiceDiscoveryComplete.
+    // Expected response: GattClientResponse.ServiceDiscoveryComplete
+    // Additional responses: zero or more GattClientResponse.ServiceDiscovered
+    // Description: Discovers all services on the remote device.
```
heinwessels-philips: `remote device` seems like a weird description. Isn't this the `connected peer`?

services/ble/Gatt.proto
```diff
@@ -94,66 +95,136 @@ enum CccdSetting
...
     rpc StartCharacteristicDiscovery(HandleRange) returns (Nothing) { option (method_id) = 3; }
-    // Discovers all descriptors within the given handle range of a characteristic.
-    // The handle range spans from the characteristic's value handle + 1 to the next characteristic's handle - 1
-    // (or the service's end handle for the last characteristic in a service).
-    // Responses: zero or more GattClientResponse.DescriptorDiscovered, followed by GattClientResponse.DescriptorDiscoveryComplete.
+
+    // Expected response: GattClientResponse.DescriptorDiscoveryComplete
+    // Additional responses: zero or more GattClientResponse.DescriptorDiscovered
+    // Description: Discovers all descriptors within the given handle range of a characteristic.
+    //   The handle range spans from the characteristic's value handle to the next characteristic's declaration handle - 1
```
heinwessels-philips, line 142: What happens if incorrect handles is passed? Will we assert? Not sure if it should be added in documentation, but is it something we can check for?

services/ble/Gatt.proto
```diff
@@ -57,13 +57,13 @@ message Service
 message Characteristic
 {
     bytes uuid = 1 [(bytes_size) = 16];
-    Handle handle = 2;
+    Handle declarationHandle = 2;
```
heinwessels-philips, line 60: This deviates from how we reference this handle everywhere else in the code base. Is this change really needed? Could the additional information not be provided through a comment?

heinwessels-philips (approved):

services/ble/Gatt.proto
danielschenk: What happens is that you will get incorrect results (you will get more handles returned of things that are not necessarily descriptors). This is a known limitation of GATT/ATT, see this Amber explanation:
image

Checking is not easy, that would mean we need to keep a local administration of characteristic discovery. That responsibility is shifted to the product itself, to keep the CN as lean as possible.

services/ble/Gatt.proto
danielschenk: We want to be as clear as possible, discussions with WARP indicated that the old name was vague and ambiguous. IMO internal code should be adjusted as much as possible to remove this ambiguity, and also, remove any hint that it could just be the only relevant handle

magi-arun (approved):

services/ble/Gatt.proto
```diff
@@ -94,66 +95,136 @@ enum CccdSetting
...
+    // Triggered by: GattClient.ReadDescriptor
+    // Description: Returns the value read from the descriptor.
     rpc ReadDescriptorResponse(ReadResult) returns (Nothing) { option (method_id) = 9; }
 
-    // Called after completion of operations with response: ReadCharacteristic, WriteCharacteristic, ReadDescriptor, WriteDescriptor
-    rpc OperationComplete(OperationResult) returns (Nothing) { option (method_id) = 10; }
+    // Triggered by: GattClient.WriteCharacteristicWithoutResponse
+    // Description: Reports whether a write without response command was successfully queued by the GATT client for transmission to the GATT server.
+    //  Success indicates the request was processed by the GATT client, but does not guarantee success on the GATT server side.
+    rpc WriteCharacteristicWithoutResponseComplete(OperationResult) returns (Nothing) { option (method_id) = 10; }
```
magi-arun: Can this response be added at end of list so that method id for existing responses do not change ?

danielschenk: Yes, will do. Although that would not solve this being a breaking change anyway (we are changing existing RPCs already)

PR #1305, opened 2026-08-11
feat: pure stateguard interface for BLE

MiquelJayson-Philips:
Created a base StateGuard that can be used on different platforms to "assert you're in one of these states". The BLE spec defines connection handles as 12-bit values which would fit in uint16_t, however they way states are determined differs per platform.

### DIFF
--- a/services/ble/StateGuard.hpp
+++ b/services/ble/StateGuard.hpp
@@ -1,8 +1,10 @@
 #ifndef SERVICES_BLE_STATE_GUARD_HPP
 #define SERVICES_BLE_STATE_GUARD_HPP
 
-#include "infra/util/MemoryRange.hpp"
+#include "infra/util/ReallyAssert.hpp"
 #include "services/ble/Gap.hpp"
+#include <algorithm>
+#include <initializer_list>
 
 namespace services
 {
@@ -15,7 +17,37 @@ namespace services
         StateGuard& operator=(const StateGuard& other) = delete;
         virtual ~StateGuard() = default;
 
-        virtual void AssertStateIs(infra::MemoryRange<const GapState> states) const = 0;
+        void AssertStateIs(std::initializer_list<GapState> states) const
+        {
+            auto currentState = DetermineCurrentState();
+
+            really_assert(std::any_of(states.begin(), states.end(), [currentState](auto state)
+                {
+                    return state == currentState;
+                }));
+        }
+
+    protected:
+        virtual GapState DetermineCurrentState() const = 0;
+    };
+
+    template<typename Owner>
+    class StateGuardWithOwner
+        : public StateGuard
+    {
+    public:
+        explicit StateGuardWithOwner(const Owner& owner)
+            : owner(owner)
+        {}
+
+    protected:
+        GapState DetermineCurrentState() const override
+        {
+            return owner.DetermineCurrentState();
+        }
+
+    private:
+        const Owner& owner;
     };
 } // namespace services
 
--- a/services/ble/StateGuard.hpp
+++ b/services/ble/StateGuard.hpp
@@ -1,6 +1,7 @@
 #ifndef SERVICES_BLE_STATE_GUARD_HPP
 #define SERVICES_BLE_STATE_GUARD_HPP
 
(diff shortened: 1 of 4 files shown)


services/ble/StateGuard.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+    class StateGuard
+    {
+    public:
+        StateGuard() = default;
+        StateGuard(const StateGuard& other) = delete;
+        StateGuard& operator=(const StateGuard& other) = delete;
+        virtual ~StateGuard() = default;
+
+        virtual void AssertStateIs(infra::MemoryRange<const GapState> states) const = 0;
```
heinwessels-philips: You could consider this for better ergonomics.
```suggestion
        virtual void AssertStateIs(std::initializer_list<const GapState> states) const = 0;
```
So that you can write `guard.AssertStateIs({ GapState::connected, GapState::advertising });`. Not sure if that's possible with a `MemoryRange`.

services/ble/StateGuard.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+    class StateGuard
+    {
+    public:
+        StateGuard() = default;
+        StateGuard(const StateGuard& other) = delete;
+        StateGuard& operator=(const StateGuard& other) = delete;
+        virtual ~StateGuard() = default;
+
+        void AssertStateIs(std::initializer_list<GapState> states) const
```
heinwessels-philips: I would add two public functions here
```C++
bool StateIs(....) { }
void AssertStateIs(...)
```
So when we have a way to report the error gracefully then we can still easily check the state.

services/ble/StateGuard.hpp
```diff
@@ -0,0 +1,35 @@
...
+        {
+            auto currentState = DetermineCurrentState();
+
+            really_assert(std::any_of(states.begin(), states.end(), [currentState](auto state)
```
heinwessels-philips: This would not provide an helpful error message. It should show the current state, and ideally the expected state, although that might be trickier because it's a list.

It's likely the customers (and us) will see this error often, so making it human readable, and providing the required info) is important.

heinwessels-philips (changes requested):

services/ble/StateGuard.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include "infra/util/ReallyAssert.hpp"
+#include "services/ble/Gap.hpp"
+#include <algorithm>
+#include <initializer_list>
+
+namespace services
+{
+
+    class StateGuard
```
heinwessels-philips, line 13: I'm also missing unit tests for this class. You can test the error message the assert function will abort with like I did in #1302

heinwessels-philips (changes requested):

services/ble/StateGuard.hpp
```diff
@@ -0,0 +1,54 @@
...
+                infra::StringOutputStream::WithStorage<16> stream;
+                for (auto it = states.begin(); it != states.end(); ++it)
+                {
+                    if (it != states.begin())
+                        stream << ",";
+                    stream << static_cast<int>(*it);
+                }
+
+                really_assert_with_msg(false,
+                    "GAP state: %d, expected: [%s]",
```
heinwessels-philips: Nice! I feel the message should put focus on that an unexpected state was found.
Also, could probably make this smaller to save some bytes.
```suggestion
                infra::StringOutputStream::WithStorage<6> stream;
                for (auto it = states.begin(); it != states.end(); ++it)
                {
                    if (it != states.begin())
                        stream << ",";
                    stream << static_cast<int>(*it);
                }

                really_assert_with_msg(false,
                    "GAP state: %d, expected: [%s]",
```

heinwessels-philips (changes requested):

services/ble/StateGuard.hpp
MiquelJayson-Philips: Adjusted, 6 would only cover 3 states, worst case 4 so changed it to 9.

heinwessels-philips (approved):

PR #1306, opened 2026-08-11
feat: allow optional providing claimable resource to ClaimingGattClientAdapter

heinwessels-philips:
The `GattClient` is often required by various different constructs, and to keep them all in sync it might be required to have a single resource shared between them all.

### DIFF
--- a/services/ble/ClaimingGattClientAdapter.cpp
+++ b/services/ble/ClaimingGattClientAdapter.cpp
@@ -10,7 +10,7 @@ namespace services
         , AttMtuExchangeObserver(attMtuExchange)
         , GapCentralObserver(gapCentral)
         , ownedResource(std::in_place)
-        , resource(ownedResource.value())
+        , resource(*ownedResource)
     {}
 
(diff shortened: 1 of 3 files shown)


services/ble/ClaimingGattClientAdapter.cpp
```diff
@@ -9,6 +9,15 @@ namespace services
         : GattClientObserver(gattClient)
         , AttMtuExchangeObserver(attMtuExchange)
         , GapCentralObserver(gapCentral)
+        , ownedResource(std::in_place)
+        , resource(ownedResource.value())
+    {}
+
+    ClaimingGattClientAdapter::ClaimingGattClientAdapter(infra::ClaimableResource& resource, GattClient& gattClient, AttMtuExchange& attMtuExchange, GapCentral& gapCentral)
```
oguzcanoguz, line 16: Is it an idea to apply the same pattern as in WithStorage, such as here WithClaimableResource?

oguzcanoguz (approved):

services/ble/ClaimingGattClientAdapter.cpp
heinwessels-philips: Decided not to go this way because it adds unnecessary code complexity to the adapter class. This is because the `WithStorage.hpp` doesn't fit this use case, and all the plumbing needs to be created. And the new constructor is perfectly readable IMO.

PR #1310, opened 2026-08-18
feat!: take address and connection parameters in GapCentral::Connect

gabrielsantosphilips:
Connect took a MAC address and an address type as separate arguments, and gave the caller no say over the link layer parameters of the connection it was about to open. The connection parameters were only reachable afterwards, through the peripheral side of the interface, so a central had to accept whatever its implementation hard-coded. That is a problem for the supervision timeout in particular, because a peripheral without a hardware ECC engine can block for well over a second while computing LE Secure Connections key material and gets supervised out by a central that assumes a short timeout.

Connect now takes the address as a single GapAddress, the requested GapConnectionParameters, and the initiating timeout. Gap.proto follows the same shape: ConnectionParameters now describes the link layer parameters, and the Connect argument is renamed to ConnectRequest.

BREAKING CHANGE: GapCentral::Connect takes a GapAddress and GapConnectionParameters instead of a MAC address and address type, and the Connect rpc argument is renamed from ConnectionParameters to ConnectRequest with initiatingTimeoutInMs moved to field 3.


services/ble/Gap.proto
```diff
@@ -392,17 +392,34 @@ message SecurityModeAndLevel
     SecurityLevelEnum level = 2;
 }
 
+// Link layer connection parameters, as defined in Bluetooth Core Specification v6.3, Vol 6, Part B, Section 4.5.1.
 message ConnectionParameters
+{
+    // Minimum and maximum connection interval, in units of 1.25 ms.
```
heinwessels-philips, line 407: Maybe we could add sensible defaults to these values? I could imagine that an customer would see this and not know what to fill it.  Similar to what we did for `AdvertisementMode`. This is something a customer requested.

Same goes for the other values in `ConnectionParameters`.

Additionally, do we need this change now to fix the bug your solving? Or could it be planned with the upcoming changes to reduce breaking changes? @MiquelJayson-Philips

heinwessels-philips (changes requested):

services/ble/Gap.proto
gabrielsantosphilips: This will be taken offline due to Philips policy.

gabrielsantosphilips:
@copilot+claude-sonnet-5:high  fix this issue in this PR creating a new commit:

❌ Linted [REPOSITORY] files with [git_diff]: Found 1 error(s) and 0 warning(s) - (0.1s) (expand for details)
  - Using [git_diff v2.54.0] https://megalinter.io/10.0.0/descriptors/repository_git_diff
  - MegaLinter key: [REPOSITORY_GIT_DIFF]
  - Rules config: identified by [git_diff]
  - CWD: [/github/workspace]
  - Command: [git diff --check]
  --Error detail:
  services/ble/test/TestGapPeripheral.cpp:67: trailing whitespace.
  +                EXPECT_EQ(param.maxConnectionIntervalMultiplier, connParam.maxConnectionIntervalMultiplier);
  services/ble/test/TestGapPeripheral.cpp:68: trailing whitespace.
  +                EXPECT_EQ(param.minConnectionIntervalMultiplier, connParam.minConnectionIntervalMultiplier);
  services/ble/test/TestGapPeripheral.cpp:69: trailing whitespace.
  +                EXPECT_EQ(param.peripheralLatency, connParam.peripheralLatency);
  services/ble/test/TestGapPeripheral.cpp:70: trailing whitespace.
  +                EXPECT_EQ(param.supervisionTimeoutMultiplier, connParam.supervisionTimeoutMultiplier);

PR #1313, opened 2026-08-18
feat(ble): specific role support for bond storage synchronizer

heinwessels-philips:
Changes the bond storage synchronizer:
- Bonds can now be stored for a specific role (peripheral or central).
- Make roles explicit between absolute storage (typically the SDK itself), and the other storage (containing more info).
- The synchronizer doesn't act on it's own (like deleting oldest bonds), except for synchronizing on construction.
- Gap implementations now need to interact through `Interactor`, which safely narrows to interface to only what's required.
- The `Interactor` does nothing by itself either (like deleting bonds). This needs to be handled by the Gap implementation, because SDK's could treat this differently.
- The `Interactor` can have it's own internal limit set on maximum bonds to store.

The intended design:
```mermaid
classDiagram
    direction TB

    class GapImpl {
        <<SDK adapter>>
    }
    class BondStorageInteractor {
        -Role role
        -uint32_t maxBondsForThisRole
    }
    class BondStorageSynchronizer {
        -uint32_t maxNumberOfBonds
        -uint32_t interactableBondStorage
    }
    class BondStorage {
        <<interface>>
    }
    class BondStorageAbsolute {
        <<interface>>
    }
    class ConfigurationStoreBondStorage {
        <<implementation>>
    }
    class SdkBondStorage {
        <<implementation>>
    }
    class ConfigurationStore {
        <<persistent storage>>
    }
    class SdkBondList {
        <<vendor BLE stack>>
    }

    GapImpl o--> BondStorageInteractor : uses
    BondStorageInteractor o--> BondStorageSynchronizer : role-scoped facade
    BondStorageSynchronizer o--> BondStorage : bondStorage
    BondStorageSynchronizer o--> BondStorageAbsolute : absoluteBondStorage
    BondStorage <|.. ConfigurationStoreBondStorage
    BondStorageAbsolute <|.. SdkBondStorage
    ConfigurationStoreBondStorage o--> ConfigurationStore
    SdkBondStorage o--> SdkBondList
```


services/ble/BondStorageInteractor.cpp
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    void BondStorageInteractor::AssertBondStoragesAreInSync()
+    {
+        bondStorageSynchroniser.AssertBondStoragesAreInSyncForRole(role);
+    }
+
+    void BondStorageInteractor::RemoveLeastRecentlyUsedBond()
+    {
+        // This removes the least recently used bond.
```
MiquelJayson-Philips: Function Name tells the same thing, comment could be removed

services/ble/BondStorageInteractor.cpp
```diff
@@ -0,0 +1,88 @@
...
+        : role(role)
+        , bondStorageSynchroniser(bondStorageSynchroniser)
+        , maxBondsForThisRole(maxBondsForThisRole)
+    {
+        bondStorageSynchroniser.AllocateInteractableBondStorage(maxBondsForThisRole);
+    }
+
+    BondStorageInteractor::BondStorageInteractor(Role role, BondStorageSynchronizer& bondStorageSynchroniser)
+        : BondStorageInteractor(role, bondStorageSynchroniser, bondStorageSynchroniser.GetMaxNumberOfBonds())
+    {
```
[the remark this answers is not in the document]
heinwessels-philips: This is the intended behaviuour.

services/ble/Gap.hpp
```diff
@@ -57,6 +57,12 @@ namespace services
         manufacturerSpecificData = 0xffu
     };
 
+    enum class Role : uint8_t
+    {
```
MiquelJayson-Philips, line 61: This is the first time I see that we use '= 0' on a enum value and leave the other values without.

Should this not be consistent, or are there special use-cases for each? GapAdvertisementDataType makes sense, but if we really care about notifying about the order should the others in Gap.hpp also not include a '=0'?

```
   enum class Role : uint8_t
    {
        central,
        peripheral,
    };

   enum class Role : uint8_t
    {
        central = 0,
        peripheral = 1,
    };

   enum class Role : uint8_t
    {
        central = 0,
        peripheral,
    };

File: services/ble/BondStorageSynchronizer.hpp [hunk 4]
```diff
@@ -18,66 +15,96 @@ namespace services
...
+        virtual uint32_t GetTotalNumberOfBonds() const = 0;
+        virtual uint32_t GetMaxNumberOfBonds() const = 0;
+        virtual std::optional<services::Bond> GetBond(Role role, const services::GapAddress& address) const = 0;
+
+        // Iteration is in least recently used order, i.e. the first bond returned is the least recently used one.
+        virtual void IterateBondedDevices(Role role, const infra::Function<void(const services::Bond&)>& onBond) = 0;
     };
 
-    class BondStorage
+    class BondStorageAbsolute
```
MiquelJayson-Philips, line 60: Not sure yet if I like the naming of BondStorageAbsolute, in particular the use of the word absolute.

It makes sense after diving deeper into the code, but at a first glance it confused me.

heinwessels-philips: I'm not completely happy with the naming of the storages either. If you have a better idea please let me know.
Alternatives for `BondStorageAbsolute
- AuthoritativeBondStorage — emphasizes it's the source of truth
- CanonicalBondStorage — the canonical/definitive record
- SdkBondStorage — explicitly the SDK-owned store
- StackBondStorage — owned by the BLE stack
- PrimaryBondStorage — the primary store the shadow mirrors
- ReferenceBondStorage — the reference against which the shadow is synced
- MinimalBondStorage — reflects that it only holds addresses (less info)
- AddressBondStorage — reflects that it's keyed/limited to addresses

Alternatives for `BondStorage`:
- ShadowBondStorage — directly names the shadow-copy role
- ExtendedBondStorage — holds extended/extra information
- DetailedBondStorage — richer per-bond detail
- EnrichedBondStorage — enriched with additional fields
- MirroredBondStorage / BondStorageMirror — mirror of the authoritative store
- LocalBondStorage — the locally maintained copy
- AnnotatedBondStorage — annotated with extra metadata

I like `AuthoritativeBondStorage`. And for the other one maybe `MirroredBondStorage` or `AnnotatedBondStorage`

.vscode/settings.json
MiquelJayson-Philips, line 1: Accidental commit?

services/ble/BondStorageSynchronizer.hpp
MiquelJayson-Philips: We are talking about two interfaces here with each their responsibilities: one holds the source of truth and other is a detailed version. These need to be synchronized. 

**I think for the one holding the source of truth:**
AuthoritativeBondStorage sounds the best option to me. Which in computer terminology is also used for e.g. DNS Server. Where an Authoritative DNS Server is the computer server that actually hold the real official DNS records. This could also work for our use case with bonds. 

**Regarding the other storage:**
I think mirror is not a good option, since it doesn't indicate there is any enrichment in terms of details, sounds like a copy of. Same goes for Annotated, how I always interpret is side notes to better understand the thing it is referring to. Personally I think EnrichedBondStorage is the best option. It indicates a small amount of metadata is added. I was also thinking about detailed option but this could mean a large amount of data is added.

MiquelJayson-Philips (changes requested):
Left some small comments, overall good job! Looks promising.

services/ble/BondStorageSynchronizer.hpp
heinwessels-philips: Renamed to `AuthoritativeBondStorage` and `EnrichedBondStorage`

services/ble/Gap.hpp
heinwessels-philips: Hmm, this is a habit I have, but can't remember why. I'll make it consistent with the rest.

services/ble/BondStorageSynchronizer.cpp
```diff
@@ -1,55 +1,131 @@
...
-    void BondStorageSynchronizerImpl::RemoveBond(hal::MacAddress address)
+    void BondStorageSynchronizerImpl::RemoveAllBondsForRole(Role role)
     {
-        referenceBondStorage.RemoveBond(address);
-        otherBondStorage.RemoveBond(address);
+        enrichedBondStorage.IterateBondedDevices(role, [this](const services::Bond& bond)
+            {
+                authoritativeBondStorage.RemoveBond(bond.address);
+            });
+        enrichedBondStorage.RemoveAllBondsForRole(role);
```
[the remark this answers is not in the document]
heinwessels-philips: This is hardening that can be added later. It should not be possible to add the same address for both roles already due to the asserts in `AddBond`.

MiquelJayson-Philips (approved):
Looks nice +1

PR #1316, opened 2026-08-19
docs(ble): note that a bond's device name can be empty

heinwessels-philips:
Note that a device name returned through the `Gap*::GetBondList` command can be empty.

### DIFF
--- a/services/ble/Gap.proto
+++ b/services/ble/Gap.proto
@@ -480,8 +480,9 @@ message Bond
     AddressWithType addressWithType = 1;
     
     // Device Name from GAP (Generic Access) service.
+    // Read from the peer on bond created and on every successful encryption change.
     // Not unique; truncated to 32 bytes.
-    // Can be empty if the device does not have a name or has not been read yet.
+    // Can be empty if read name is empty, or has not been read yet.
     bytes deviceName = 2 [(bytes_size) = 32];
 }
 
--- a/services/ble/Gap.proto
+++ b/services/ble/Gap.proto
@@ -482,7 +482,7 @@ message Bond
     // Device Name from GAP (Generic Access) service.
     // Read from the peer on bond created and on every successful encryption change.
     // Not unique; truncated to 32 bytes.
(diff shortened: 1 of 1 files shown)


services/ble/Gap.proto
```diff
@@ -476,8 +476,13 @@ message ResolvedPrivateAddress
...
 message Bond
 {
-    AddressWithType addressWithType = 1;        // Identity address
-    bytes deviceName = 2 [(bytes_size) = 32];   // Device Name from GAP (Generic Access) service. Not unique; truncated to 32 bytes.
+    // Identity address of the peer
+    AddressWithType addressWithType = 1;
+    
+    // Device Name from GAP (Generic Access) service.
+    // Not unique; truncated to 32 bytes.
+    // Can be empty if the device does not have a name or has not been read yet.
```
oguzcanoguz: 'device has an empty name'
Because a device has to have this field.

services/ble/Gap.proto
```diff
@@ -476,8 +476,13 @@ message ResolvedPrivateAddress
 
 message Bond
 {
-    AddressWithType addressWithType = 1;        // Identity address
-    bytes deviceName = 2 [(bytes_size) = 32];   // Device Name from GAP (Generic Access) service. Not unique; truncated to 32 bytes.
+    // Identity address of the peer
+    AddressWithType addressWithType = 1;
+    
+    // Device Name from GAP (Generic Access) service.
```
oguzcanoguz, line 482: So
1. Name is read and empty.
2. Name is not read yet.
  a. Connected: Name is being read
  b. Disconnected: Name could not be read before disconnecting.

1 and 2.a are fine. 
Can 2.b happen, that is, do we write to configstore before reading the name?
If so do we plan to have a mechanism to recover from it in the next connection?

oguzcanoguz (approved):

services/ble/Gap.proto
heinwessels-philips: Applied

PR #1320, opened 2026-08-25
chore: align GapPeripheral service formatting with GapCentral service in Gap.proto 

MiquelJayson-Philips:
Reformatted GapPeripheral and GapPeripheralResponse service comments to match the GapCentral documentation style: added structured labels (Expected response:, Description:, Triggered by:), indented continuation lines, and clarified the SetAdvertisementData/SetScanResponseData state behavior to resolve a contradiction between Allowed states and the description.


services/ble/Gap.proto
```diff
@@ -259,74 +259,76 @@ service GapPeripheral
...
+    // Expected response: GapPeripheralResponse.CurrentState with state standby
     rpc Standby(Nothing) returns (Nothing) { option (method_id) = 3; }
 
     // Allowed states: standby
-    // Maximum advertisement data length for this method is 28 bytes. Server reserves an additional 3 bytes for flags data with the flag bits 'LE General Discoverable Mode' and 'BR/EDR Not Supported' enabled.
-    // See Core Specification Supplement, Part A, Section 1.3
-    // Server will not check validity of the advertisement data and only maximum length of each field will be passed to BLE stack
-    // If set, the advertisement data remains in effect until it is overwritten or the device is rebooted.
-    // If not set by this method, server will have as advertisement data only flags field.
+    // Description: Does not assert outside standby; changes take effect on the next GapPeripheral.Advertise call.
```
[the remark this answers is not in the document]
MiquelJayson-Philips: These are the only two functions where calling it will not result in assertion, since it is only used for storing data locally. These values will be used in the next GapPeripheral Advertise call. This has its own states that are asserting.

The intention to keep this in only in "allowed states: standby" is to first align with advertising and second to prevent wrong usage of the APIs.

services/ble/Gap.proto
```diff
@@ -259,74 +259,76 @@ service GapPeripheral
...
-    // Results in GapPeripheralResponse.CurrentState with current state
+    // Expected response: GapPeripheralResponse.CurrentState with current state
     rpc GetState(Nothing) returns (Nothing) { option (method_id) = 1; }
 
     // Allowed states: standby
-    // When an advertisement mode is not supported by a platform it results in an assertion
-    // If an advertisement mode is not supported by a configuration then it results in an assertion
-    // Results in GapPeripheralResponse.CurrentState with state advertising
+    // Expected response: GapPeripheralResponse.CurrentState with state advertising
+    // Description: When an advertisement mode is not supported by a platform it results in an assertion.
```
magi-arun: Is there a reason why we have both lines which seems same

services/ble/Gap.proto
```diff
@@ -259,74 +259,76 @@ service GapPeripheral
...
+    //   If an advertisement mode is not supported by a configuration then it results in an assertion.
     rpc Advertise(AdvertisementMode) returns (Nothing) { option (method_id) = 2; }
 
     // Allowed states: standby, advertising, connected
-    // Results in GapPeripheralResponse.CurrentState with state standby
+    // Expected response: GapPeripheralResponse.CurrentState with state standby
     rpc Standby(Nothing) returns (Nothing) { option (method_id) = 3; }
 
     // Allowed states: standby
```
magi-arun: Is this line needed now since it is allowed in all states

services/ble/Gap.proto
magi-arun: If we remove the allowed state then we may not need to mention about assert here

services/ble/Gap.proto
```diff
@@ -259,74 +259,76 @@ service GapPeripheral
...
-    // Server will not check validity of the scan response data and only maximum length of each field will be passed to BLE stack
-    // If set, the scan response data remains in effect until it is overwritten or the device is rebooted.
-    // If not set by this method, server will have as scan response data only service uuids.
+    // Description: Does not assert outside standby; changes take effect on the next GapPeripheral.Advertise call.
+    //   Maximum scan response length for this method is 13 bytes. Server reserves 18 bytes for streaming service uuids. Scan Response is only possible with scannable advertisements
+    //   Server will not check validity of the scan response data and only maximum length of each field will be passed to BLE stack
+    //   If set, the scan response data remains in effect until it is overwritten or the device is rebooted.
+    //   If not set by this method, server will have as scan response data only service uuids.
     rpc SetScanResponseData(ScanResponseData) returns (Nothing) { option (method_id) = 5; }
```
magi-arun, line 287: Same as above

magi-arun (approved):

services/ble/Gap.proto
MiquelJayson-Philips: Good catch, removed the duplicated line!

services/ble/Gap.proto
MiquelJayson-Philips: Removed it to prevent confusion.

services/ble/Gap.proto
MiquelJayson-Philips: Removed talk about assertion and allowed states.

services/ble/Gap.proto
MiquelJayson-Philips: I removed allowed state and mention that about not asserting

services/ble/Gap.proto
```diff
@@ -259,74 +259,73 @@ service GapPeripheral
...
     rpc Standby(Nothing) returns (Nothing) { option (method_id) = 3; }
 
-    // Allowed states: standby
-    // Maximum advertisement data length for this method is 28 bytes. Server reserves an additional 3 bytes for flags data with the flag bits 'LE General Discoverable Mode' and 'BR/EDR Not Supported' enabled.
-    // See Core Specification Supplement, Part A, Section 1.3
-    // Server will not check validity of the advertisement data and only maximum length of each field will be passed to BLE stack
-    // If set, the advertisement data remains in effect until it is overwritten or the device is rebooted.
-    // If not set by this method, server will have as advertisement data only flags field.
+    // Description: Changes take effect on the next GapPeripheral.Advertise call.
+    //   Maximum advertisement data length for this method is 28 bytes. Server reserves an additional 3 bytes for flags data with the flag bits 'LE General Discoverable Mode' and 'BR/EDR Not Supported' enabled.
```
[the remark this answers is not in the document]
MiquelJayson-Philips: Intentionally removed

PR #1326, opened 2026-09-02
feat(mbedTLS): refactored, upgraded and removed deprecated mbedTLSv3 calls

Petzep:
This pull request updates the certificate handling code to use the PSA Crypto API for private and public key operations, improving security and maintainability. It also adds a utility method to the ASN.1 formatter for easier DER encoding insertion.

**Certificate key handling improvements:**

* Refactored `CertificatesMbedTls::WritePrivateKey` to use PSA Crypto API for exporting private keys instead of manual ASN.1 formatting, ensuring better integration with hardware-backed key storage and improved security.
* Updated public key encoding in certificate generation to use `mbedtls_pk_write_pubkey_der` and the new `AddDer` method, replacing manual ASN.1 construction for SubjectPublicKeyInfo.
* Ensured PSA Crypto is initialized during certificate object construction by calling `psa_crypto_init()`.

**Infrastructure and utility enhancements:**

* Added `AddConstruct` method to `infra::Asn1Formatter` to allow direct insertion of constructed data into ASN.1 streams. [[1]](diffhunk://#diff-fa7c93b0d7beb023c1a507aa472edc36d1682c13baf2012662f8175600017bd8R133-R137) [[2]](diffhunk://#diff-53fe8aa7b346843053980aec49f4ddff7fa49bf98cf0d3adea746f6a77b6e939R30)

**Code cleanup:**

* Removed unused `mbedtls/oid.h` include from `CertificatesMbedTls.cpp`.

### DIFF
--- a/infra/syntax/Asn1Formatter.cpp
+++ b/infra/syntax/Asn1Formatter.cpp
@@ -130,6 +130,11 @@ namespace infra
         stream << infra::text << 'Z';
     }
 
+    void Asn1Formatter::AddDer(infra::ConstByteRange der)
+    {
+        stream << der;
+    }
+
     Asn1ContainerFormatter Asn1Formatter::StartSequence()
     {
         stream << static_cast<uint8_t>(Tag::Constructed | Tag::Sequence);
--- a/infra/syntax/Asn1Formatter.cpp
+++ b/infra/syntax/Asn1Formatter.cpp
@@ -130,9 +130,10 @@ namespace infra
         stream << infra::text << 'Z';
     }
 
-    void Asn1Formatter::AddDer(infra::ConstByteRange der)
+    void Asn1Formatter::AddConstructed(infra::ConstByteRange constructed)
     {
-        stream << der;
+        really_assert(!constructed.empty() && (constructed.front() & Tag::Constructed) != 0);
+        stream << constructed;
     }
 
     Asn1ContainerFormatter Asn1Formatter::StartSequence()
--- a/infra/syntax/Asn1Formatter.cpp
+++ b/infra/syntax/Asn1Formatter.cpp
@@ -1,4 +1,5 @@
 #include "infra/syntax/Asn1Formatter.hpp"
+#include "infra/util/ReallyAssert.hpp"
 
 namespace
 {
--- a/services/network/CertificatesMbedTls.cpp
+++ b/services/network/CertificatesMbedTls.cpp
@@ -2,8 +2,8 @@
 #include "infra/stream/ByteOutputStream.hpp"
 #include "infra/stream/StringOutputStream.hpp"
 #include "infra/util/ReallyAssert.hpp"
-#include "mbedtls/build_info.h"
 #include "mbedtls/pk.h"
+#include "mbedtls/version.h"
 #include "services/util/MbedTlsRandomDataGeneratorWrapper.hpp"
 
 #if MBEDTLS_VERSION_MAJOR < 3
--- a/services/network/CertificatesMbedTls.cpp
+++ b/services/network/CertificatesMbedTls.cpp
@@ -2,8 +2,8 @@
 #include "infra/stream/ByteOutputStream.hpp"
 #include "infra/stream/StringOutputStream.hpp"
 #include "infra/util/ReallyAssert.hpp"
+#include "mbedtls/build_info.h"
 #include "mbedtls/pk.h"
-#include "mbedtls/version.h"
 #include "services/util/MbedTlsRandomDataGeneratorWrapper.hpp"
 
 #if MBEDTLS_VERSION_MAJOR < 3
@@ -121,6 +121,8 @@ namespace services
         stream << infra::AsBase64(infra::ConstByteRange(contents, contents + contentsSize));
         stream << "-----END RSA PRIVATE KEY-----\r\n";
         stream << '\0';
+
+        mbedtls_platform_zeroize(contents, sizeof(contents));
     }
 
     void CertificatesMbedTls::WriteOwnCertificate(infra::BoundedString& outputBuffer, hal::SynchronousRandomDataGenerator& randomDataGenerator)
--- a/services/network/CertificatesMbedTls.cpp
+++ b/services/network/CertificatesMbedTls.cpp
@@ -3,6 +3,7 @@
 #include "infra/stream/StringOutputStream.hpp"
 #include "infra/util/ReallyAssert.hpp"
 #include "mbedtls/pk.h"
+#include "mbedtls/platform_util.h"
 #include "mbedtls/version.h"
 #include "services/util/MbedTlsRandomDataGeneratorWrapper.hpp"
 
--- a/services/network/CertificatesMbedTls.cpp
+++ b/services/network/CertificatesMbedTls.cpp
@@ -105,11 +105,12 @@ namespace services
     {
         psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
         really_assert(mbedtls_pk_get_psa_attributes(&privateKey, PSA_KEY_USAGE_SIGN_HASH, &attributes) == 0);
+        really_assert(mbedtls_pk_get_type(&privateKey) == MBEDTLS_PK_RSA);
 
         mbedtls_svc_key_id_t keyId = MBEDTLS_SVC_KEY_ID_INIT;
         really_assert(mbedtls_pk_import_into_psa(&privateKey, &attributes, &keyId) == 0);
 
-        unsigned char contents[2048];
+        unsigned char contents[PSA_EXPORT_KEY_PAIR_MAX_SIZE];
         std::size_t contentsSize = 0;
         auto status = psa_export_key(keyId, contents, sizeof(contents), &contentsSize);
         psa_destroy_key(keyId);
@@ -178,7 +179,7 @@ namespace services
                         auto length = mbedtls_pk_write_pubkey_der(&privateKey, spki, sizeof(spki));
                         really_assert(length > 0);
 
-                        tbsSequence.AddDer(infra::ConstByteRange(spki + sizeof(spki) - length, spki + sizeof(spki)));
+                        tbsSequence.AddConstructed(infra::ConstByteRange(spki + sizeof(spki) - length, spki + sizeof(spki)));
                     }
 
                     //  v3 Extensions
--- a/services/network/CertificatesMbedTls.cpp
+++ b/services/network/CertificatesMbedTls.cpp
@@ -112,9 +112,8 @@ namespace services
 
         unsigned char contents[PSA_EXPORT_KEY_PAIR_MAX_SIZE];
         std::size_t contentsSize = 0;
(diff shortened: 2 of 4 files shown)


services/network/CertificatesMbedTls.cpp
```diff
@@ -102,28 +103,22 @@ namespace services
...
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(DQ)));
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(QP)));
-        }
+        psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+        really_assert(mbedtls_pk_get_psa_attributes(&privateKey, PSA_KEY_USAGE_SIGN_HASH, &attributes) == 0);
+
+        mbedtls_svc_key_id_t keyId = MBEDTLS_SVC_KEY_ID_INIT;
+        really_assert(mbedtls_pk_import_into_psa(&privateKey, &attributes, &keyId) == 0);
+
+        unsigned char contents[2048];
```
danielschenk: This is quite some data to allocate on the stack. I think we can easily avoid this by passing the `outputBuffer` directly to `psa_export_key` (after line 120) and then advancing the stream manually with the written number of bytes.

infra/syntax/Asn1Formatter.cpp
```diff
@@ -130,6 +130,11 @@ namespace infra
         stream << infra::text << 'Z';
     }
 
+    void Asn1Formatter::AddDer(infra::ConstByteRange der)
+    {
+        stream << der;
```
danielschenk: This is just an open door to append any sequence of bytes to the stream. Can't we shift the logic into this class? Eg. something like:
```cpp
void Asn1Formatter::AddPublicKeyAsDer(const <INSERT TYPE HERE>& privateKey)
{
    // ... stuff currently in `WriteOwnCertificate` ...
```

(I'm not into the subject matter very deep, maybe @richardapeters has a better idea). With that we could also solve the other large stack allocation, because I assume this formatter could make mbedTLS write directly into the right place (see other comment).

danielschenk (changes requested):

services/network/CertificatesMbedTls.cpp
danielschenk, line 1: General comment: is it possible to add unit tests for any of this? (Applies to whole PR)

services/network/CertificatesMbedTls.cpp
richardapeters: So if this block of code would be inserted in the stream handling below, that could look like this:

std::size_t contentsSize = 0;
auto oldSize = outputBuffer.size();
outputBuffer.resize(outputBuffer.max_size());
auto status = psa_export_key(keyId, outputBuffer.begin() + oldSize, outputBuffer.size() - oldSize, &contentsSize);
psa_destroy_key(keyId);
really_assert(status == PSA_SUCCESS);
outputBuffer.resize(oldSize + contentsSize);

The stream does not need adjusting for this

services/network/CertificatesMbedTls.cpp
Petzep: `write_private_key` in [services/network/test/TestCertificatesMbedTls.cpp](https://github.com/Petzep/amp-embedded-infra-lib/blob/main/services/network/test/TestCertificatesMbedTls.cpp) covers the key generation

I'll add a test in `infra/syntax/test/TestAsn1Formatter.cpp` to test the `AddDer` function.

infra/syntax/Asn1Formatter.cpp
richardapeters: The Asn1Formatter does not know anything about public keys, but if I look at the code replaced below, then what is inserted is an ASN1 sequence produced by mbedtls_pk_write_pubkey_der. I have thought about AddSequence instead of AddDer, but the der inserted already contains the header indicating that it's a sequence. Perhaps AddConstructed? Because this function can also be used to insert a Set or a ContextSpecific, which each carry ASN1's Constructed tag

Petzep: If the function name is changed to `AddConstructed`, can should assert on `(der.front() & Tag::Constructed) != 0` to limit unintentional use?

services/network/CertificatesMbedTls.cpp
Petzep: The psa_export_key() output still needs to be converted to Base64.

Having said that I think we are stuck with putting the data on the stack first.
The generated key size in the test is around 1.2k, nevertheless the chosen 2048 is a magic number. PSA_EXPORT_KEY_PAIR_MAX_SIZE  should have been used. But looking more in-dept I noticed that PSA_EXPORT_KEY_PAIR_MAX_SIZE evaluates to 2364 bytes. (on the host system).

Is the stack usage in this function unjustified or is it just undesirable?

infra/syntax/Asn1Formatter.cpp
```diff
@@ -130,6 +130,12 @@ namespace infra
         stream << infra::text << 'Z';
     }
 
+    void Asn1Formatter::AddConstructed(infra::ConstByteRange constructed)
+    {
+        really_assert(!constructed.empty() && (constructed.front() & Tag::Constructed) != 0);
```
richardapeters, line 136: Ah, I like this assert!

richardapeters (approved):

services/network/CertificatesMbedTls.cpp
```diff
@@ -102,28 +103,22 @@ namespace services
...
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(E)));
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(D)));
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(P)));
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(Q)));
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(DP)));
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(DQ)));
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(QP)));
-        }
+        psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+        really_assert(mbedtls_pk_get_psa_attributes(&privateKey, PSA_KEY_USAGE_SIGN_HASH, &attributes) == 0);
```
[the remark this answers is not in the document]
Petzep: get attirbutes does not allow for OR'ed flags.
The suggestion will break the compilation. Besides that, it was also unnecessary: the function already unconditionally adds PSA_KEY_USAGE_EXPORT (and COPY) to the resulting attributes internally (more_usage |= PSA_KEY_USAGE_EXPORT | PSA_KEY_USAGE_COPY;), so exporting still would have worked with just PSA_KEY_USAGE_SIGN_HASH, which is what the code used before that commit.

services/network/CertificatesMbedTls.cpp
```diff
@@ -102,30 +103,26 @@ namespace services
...
+        really_assert(mbedtls_pk_get_type(&privateKey) == MBEDTLS_PK_RSA);
+
+        mbedtls_svc_key_id_t keyId = MBEDTLS_SVC_KEY_ID_INIT;
+        really_assert(mbedtls_pk_import_into_psa(&privateKey, &attributes, &keyId) == 0);
```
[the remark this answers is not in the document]
Petzep: Will not be implemented
https://arm-software.github.io/CMSIS-mbedTLS/latest/group__attributes.html#ga0ec645e1fdafe59d591104451ebf5680

The reset is optional and the attributes hold no security critical information

services/network/CertificatesMbedTls.cpp
```diff
@@ -176,24 +173,14 @@ namespace services
...
+                        constexpr std::size_t subjectPublicKeyInfoOverhead = 24;
+                        unsigned char spki[PSA_EXPORT_PUBLIC_KEY_MAX_SIZE + subjectPublicKeyInfoOverhead];
 
-                        {
-                            auto publicKeyBitString = publicKeyInfoSequence.StartBitString();
-                            {
-                                auto rsaPublicKeySequence = publicKeyBitString.StartSequence();
-                                mbedtls_rsa_context* rsaContext = mbedtls_pk_rsa(privateKey);
-                                really_assert(rsaContext != nullptr);
+                        auto length = mbedtls_pk_write_pubkey_der(&privateKey, spki, sizeof(spki));
```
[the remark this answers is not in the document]
Petzep: The overhead is sequence overhead, not psa overhead

services/network/CertificatesMbedTls.cpp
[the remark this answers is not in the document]
Petzep: This is incorrect. MBEDTLS_USE_PSA_CRYPTO enables PSA for the internal libraries

services/network/CertificatesMbedTls.cpp
```diff
@@ -176,24 +174,14 @@ namespace services
...
+                        really_assert(length > 0);
 
-                                rsaPublicKeySequence.AddBigNumber(MakeByteRange(rsaContext->MBEDTLS_PRIVATE(N)));
-                                rsaPublicKeySequence.AddBigNumber(MakeByteRange(rsaContext->MBEDTLS_PRIVATE(E)));
-                            }
-                        }
+                        tbsSequence.AddConstructed(infra::ConstByteRange(spki + sizeof(spki) - length, spki + sizeof(spki)));
```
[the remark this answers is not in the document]
Petzep: AddConstructed is tested by /workspaces/amp-embedded-infra-lib/infra/syntax/test/TestAsn1Formatter.cpp

services/network/CertificatesMbedTls.cpp
```diff
@@ -102,30 +104,26 @@ namespace services
...
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(DP)));
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(DQ)));
-            sequence.AddBigNumber(MakeByteRange(rsaContext.MBEDTLS_PRIVATE(QP)));
-        }
+        psa_key_attributes_t attributes = PSA_KEY_ATTRIBUTES_INIT;
+        really_assert(mbedtls_pk_get_psa_attributes(&privateKey, PSA_KEY_USAGE_SIGN_HASH, &attributes) == 0);
+        really_assert(mbedtls_pk_get_type(&privateKey) == MBEDTLS_PK_RSA);
+
+        mbedtls_svc_key_id_t keyId = MBEDTLS_SVC_KEY_ID_INIT;
+        really_assert(mbedtls_pk_import_into_psa(&privateKey, &attributes, &keyId) == 0);
```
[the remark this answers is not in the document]
Petzep: mbedTLSv2-LTS supports PSA

danielschenk (approved):

PR #1327, opened 2026-09-02
fix: sesame/echo receive deadlock on sliced messages

andjordan:
## What goes wrong

An Echo link running over `SesameSecured` can stop receiving. Sending keeps working, but the receive window is never credited back, so the peer eventually runs out of window and the link is stuck for good.

Two things have to coincide, which is why it is rare: a message that spans several Sesame messages, and the next Sesame message already queued when the previous one is released.

## Why

`SesameSecured` kept the lower-layer reader inside the object owned by its `SharedOptional`, so that reader was released from within `SharedOptional::Destruct()`:

```
EchoOnStreams::ReleaseReader() -> readerPtr = nullptr
 -> SharedOptional<ReceiveBufferReader>::Destruct        [allocatable still false]
   -> ~ReceiveBufferReader -> ~SharedPtr(member `reader`) -> SharedPtr::Reset(nullptr, nullptr)
     -> AccessedBySharedPtr::Deallocate -> SesameWindowed's reader-release action
       -> TryForwardReceivedMessage -> SesameSecured::ReceivedMessage
         -> receiveBufferReader.Emplace(...)             <-- allocatable == false
```

`Emplace` asserts on that, but the assert is compiled out in release builds, so it simply carries on:

1. `std::optional::emplace` constructs the **new** reader in the storage of the object whose destructor is still unwinding. Its `reader` member takes a reference on `SesameWindowed::readerAccess`.
2. The stack unwinds back into the **old** `~SharedPtr::Reset`, which ends with `control = nullptr; object = nullptr` — wiping the new object's `reader` member.
3. That reference is now never released. `readerAccess` stays referenced forever, its action never runs, and `SesameWindowed` neither credits the window nor forwards anything again.

## Fix

Keep the lower-layer reader in a `SesameSecured` member and drop it from the `NotifyingSharedOptional` `onAllocatable` callback. That runs in `Deallocate()`, after `allocatable` is back to `true` and the old object is fully gone — the same shape as `AccessedBySharedPtr`, where the action is the last step and re-entering is safe.

`ReleaseReceivedReader()` moves the member into a local before dropping it. Dropping it directly would let a re-entrant `ReceivedMessage()` assign the member, only for the outer `SharedPtr::Reset` to wipe it again on the way out — the same bug one level up.

## Test

`sliced_messages_are_received_while_a_method_is_still_executing` in `TestEchoInstantiationSecured.cpp`. A method defers its `MethodDone()` while several multi-slice messages are queued from the peer, so the first slice of the next message lands while the deserializer is still busy. Without the fix it trips the `allocatable` assert; in a build without asserts the remaining messages simply never arrive.

`SesameSecured` is needed to reproduce — without it `EchoOnStreams` holds the `AccessedBySharedPtr` reader directly, and re-entering there is safe.


services/util/SesameSecured.cpp
```diff
@@ -171,7 +171,16 @@ namespace services
...
+        receivedReader = std::move(reader);
+        Sesame::GetObserver().ReceivedMessage(receiveBufferReader.Emplace(receiveBuffer));
+    }
+
+    void SesameSecured::ReleaseReceivedReader()
+    {
+        // Releasing the reader lets the lower layer forward the next message into receiveBufferReader.
+        // Move it out first, so that a re-entrant ReceivedMessage() assigns to an empty receivedReader
+        // instead of having its assignment overwritten while this SharedPtr is being reset.
+        auto reader = std::move(receivedReader);
```
andjordan:
```suggestion
        receivedReader = nullptr;
```

richardapeters (approved):

PR #1332, opened 2026-09-08
fix(infra/util): allow BoundedVector::swap with a full vector

### LINKED ISSUE #1314: BoundedVector::swap() incorrectly rejects swapping with a full vector
### infra::BoundedVector::swap() currently contains the following assertion:
assert(size() <= other.max_size() && other.size() < max_size());
The second condition uses < instead of <=.
This prevents valid swaps when other is full and both vectors have sufficient capacity.

### **Minimal reproduction**
#include "infra/util/BoundedVector.hpp"
int main()
{
    infra::BoundedVector<int>::WithMaxSize<2> a;
    infra::BoundedVector<int>::WithMaxSize<2> b;

    a.push_back(1);
    a.push_back(2);

    b.push_back(3);
    b.push_back(4);

    a.swap(b);
}
Both vectors have a maximum size of 2 and both contain exactly 2 elements.
However, the assertion evaluates to:
other.size() < max_size()
2 < 2
which is false.
As a result, the swap triggers an assertion failure in debug builds.

### **Expected behavior**
Two BoundedVector instances with sufficient capacity should be able to swap even when one or both vectors are full.
For vectors with the same capacity, swapping two full vectors should be valid.

### **Suggested fix**
The assertion appears to need:
assert(size() <= other.max_size() && other.size() <= max_size());
This would allow the full-vector case while still checking that each vector has enough capacity for the other's contents.

### **Additional note**
The current condition is also asymmetric: swapping can be rejected depending on which vector is used as the left-hand side when capacities differ.
I believe this should be covered by a unit test for BoundedVector::swap().

CaptainAni187:
Fixes #1314

`BoundedVector::swap()` guards capacity with

```cpp
assert(size() <= other.max_size() && other.size() < max_size());
```

The second condition uses `<` where the first uses `<=`. Swapping is therefore rejected whenever the right-hand vector is exactly full, even though both vectors have enough room for the other's contents.

Two consequences:

- two full vectors of equal capacity can never be swapped
- the check is asymmetric — `full.swap(smaller)` passes while `smaller.swap(full)` aborts, so whether a valid swap succeeds depends on which vector is the left-hand side

The reproduction from the issue aborts on `main`:

```cpp
infra::BoundedVector<int>::WithMaxSize<2> a, b;
a.push_back(1); a.push_back(2);
b.push_back(3); b.push_back(4);
a.swap(b);   // Assertion failed: other.size() < max_size()  ->  2 < 2
```

The body below the assertion already handles this case correctly — the three loops copy the shared prefix and move the tail either way — so only the guard needed changing.

## Change

Use `<=` in both halves of the assertion. Each vector is still required to have capacity for the other's contents; only the off-by-one is removed.

## Tests

Two tests added alongside the existing swap tests:

- `TestSwapTwoFullVectors` — the reported case, both vectors full at equal capacity
- `TestSwapWithFullVector` — a smaller vector swapped with a full one, the direction the asymmetry rejected

`TestBoundedVector.cpp` passes 48/48 with this change. With the assertion reverted, the suite aborts at `TestSwapTwoFullVectors`, so the tests pin the guard rather than the surrounding logic.

I ran that file directly against GoogleTest rather than through the CMake presets — `cmake --preset host` stalls while fetching dependencies on my machine, so I could not run the full `host` build locally. Everything here is header-only container code, so CI should be a clean signal.

### DIFF
--- a/infra/util/BoundedVector.hpp
+++ b/infra/util/BoundedVector.hpp
@@ -550,7 +550,7 @@ namespace infra
     {
         using std::swap;
 
-        assert(size() <= other.max_size() && other.size() < max_size());
+        assert(size() <= other.max_size() && other.size() <= max_size());
 
         for (size_type i = 0; i < size() && i < other.size(); ++i)
         {
--- a/infra/util/test/TestBoundedVector.cpp
+++ b/infra/util/test/TestBoundedVector.cpp
@@ -408,6 +408,36 @@ TEST(BoundedVectorTest, TestSwapSmallerWithLarger)
     EXPECT_EQ(expectedVector2, vector2);
 }
 
+TEST(BoundedVectorTest, TestSwapTwoFullVectors)
+{
+    int range1[3] = { 0, 1, 2 };
+    infra::BoundedVector<int>::WithMaxSize<3> vector1(range1, range1 + 3);
+    int range2[3] = { 3, 4, 5 };
+    infra::BoundedVector<int>::WithMaxSize<3> vector2(range2, range2 + 3);
+
+    infra::swap(vector1, vector2);
+
+    infra::BoundedVector<int>::WithMaxSize<3> expectedVector1(range2, range2 + 3);
+    infra::BoundedVector<int>::WithMaxSize<3> expectedVector2(range1, range1 + 3);
+    EXPECT_EQ(expectedVector1, vector1);
+    EXPECT_EQ(expectedVector2, vector2);
+}
+
+TEST(BoundedVectorTest, TestSwapWithFullVector)
+{
+    int range1[1] = { 0 };
+    infra::BoundedVector<int>::WithMaxSize<3> vector1(range1, range1 + 1);
+    int range2[3] = { 1, 2, 3 };
+    infra::BoundedVector<int>::WithMaxSize<3> vector2(range2, range2 + 3);
+
+    infra::swap(vector1, vector2);
+
+    infra::BoundedVector<int>::WithMaxSize<3> expectedVector1(range2, range2 + 3);
+    infra::BoundedVector<int>::WithMaxSize<3> expectedVector2(range1, range1 + 1);
+    EXPECT_EQ(expectedVector1, vector1);
+    EXPECT_EQ(expectedVector2, vector2);
+}
+
 TEST(BoundedVectorTest, TestClear)
 {
     infra::BoundedVector<int>::WithMaxSize<5> vector(std::size_t(2), 4);


CaptainAni187:
Note on the four red checks — all four are fork-permission failures rather than anything in this change:

- `SonarCloud`: `ERROR Not authorized or project not found. Please check the 'SONAR_TOKEN' environment variable`
- `dependency-submission`: `HTTP 403 ... Resource not accessible by integration` posting the dependency-graph snapshot
- `fuzz_pr (address)` and `(undefined)`: CIFuzz fails cloning `amp-embedded-infra-lib-metrics.git` (exit 128)

Each needs a secret or write scope that a fork PR doesn't get. PRs #1327–#1331 pass all four, and those were branched inside this repo rather than from a fork.

The build and test jobs are green: Host Build & Test on ubuntu-latest, macos-latest, clang-msvc and amp-devcontainer-cpp, plus Host Build without MbedTLS, Embedded Build, Linting and CodeQL.

Happy to rebase or move the branch if it's easier to run the full pipeline from an in-repo branch.

richardapeters (approved):

richardapeters:
@CaptainAni187 Thank you for your contribution! Indeed, due to forking some of our jobs cannot run. We haven't found a secure solution to that yet, since there is a combination of compiling code from an unknown source (the fork) and executing the resulting code. So for now, those four jobs are acceptable to be failing for remote pull requests

CaptainAni187:
Thanks for merging, and that makes sense, building and running code from a fork with secrets in scope is the hard part.

One small thing you may already have considered: `fuzzing-pr.yml` line 14 skips fuzzing for dependabot, with the comment "dependabot does not have access to secrets". Fork PRs are in the same position, so extending that to `if: ${{ github.actor != 'dependabot[bot]' && !github.event.pull_request.head.repo.fork }}` would make those two jobs skip rather than go red. It does not solve anything, but it keeps the failures on a fork PR down to the ones that are actually informative.

`dependency-submission` is the one where the split is clean, since nothing there needs to execute in the privileged context. Generating the snapshot in the `pull_request` job and uploading it as an artifact, then submitting it from a `workflow_run` job that has `contents: write`, keeps the untrusted side unprivileged, and only JSON crosses the boundary.

Sonar is harder because the coverage build genuinely has to run the code. If you ever want it on forks, a GitHub Environment with required reviewers on the sonar job is the usual route, since it releases `SONAR_TOKEN` only after someone has looked at the diff.

Happy to put any of that up as a PR if it would be useful, otherwise feel free to ignore.

PR #1337, opened 2026-09-11
feat(lwip): enhanced mbedTLS and lwIP with vendor flexibility and dependency removal

Petzep:
This pull request introduces several improvements and bug fixes to the networking codebase, primarily focused on enhancing robustness when the default network interface (`netif_default`) is not set, improving IPv6 zone handling, and updating compatibility with newer mbedTLS versions. The changes are mostly defensive checks, utility refactoring, and minor API adjustments.

**Robustness and Defensive Programming:**

* Added checks for `netif_default` being `nullptr` in various methods across `LightweightIp`, `ArpMacResolverLwIp`, and `Nd6MacResolverLwIp`, preventing potential null pointer dereferences and returning safe defaults when the network interface is not available. [[1]](diffhunk://#diff-1d59cdc716e0d17214c3e699a0b0a0938737f039330d7313a3c0baa22770cc2dR96-R114) [[2]](diffhunk://#diff-776fc22efec770dafae343dce0e878645171673eeca07e0bfa74e0ccccb7cbfbR105-R107) [[3]](diffhunk://#diff-776fc22efec770dafae343dce0e878645171673eeca07e0bfa74e0ccccb7cbfbR166-R168)

**IPv6 Zone Handling Improvements:**

* Introduced a new utility function `DefaultNetifZone()` to safely retrieve the IPv6 zone from `netif_default`, and refactored all direct accesses to use this function, improving code safety and maintainability. [[1]](diffhunk://#diff-d5b3026f22cb991ce3b5c14bb1c20a042eb8af787e2dce08f48eb1596f53638aR29-R33) [[2]](diffhunk://#diff-d5b3026f22cb991ce3b5c14bb1c20a042eb8af787e2dce08f48eb1596f53638aL86-R91) [[3]](diffhunk://#diff-d5b3026f22cb991ce3b5c14bb1c20a042eb8af787e2dce08f48eb1596f53638aL116-R121) [[4]](diffhunk://#diff-d5b3026f22cb991ce3b5c14bb1c20a042eb8af787e2dce08f48eb1596f53638aL279-R284)

**Platform and Build System Updates:**

* Updated conditional compilation in `LightweightIp.cpp` to use `#if NO_SYS` instead of `#ifndef ESP_PLATFORM`, aligning with standard lwIP configuration macros.
* Updated mbedTLS header includes to use `mbedtls/version.h` instead of `mbedtls/build_info.h` for compatibility with mbedTLS v3 and above.

**API and Weak Symbol Usage:**

* Marked `mbedtls_hardware_poll` as a weak symbol with `__weak`, allowing it to be overridden in user code or platform-specific implementations.

**TCP Parameter Handling:**

* Removed ESP-specific TCP buffer and window size definitions from `ConnectionLwIp.hpp`, now always using the lwIP defaults for consistency across platforms.

### DIFF
--- a/lwip/lwip_cpp/DatagramLwIp.cpp
+++ b/lwip/lwip_cpp/DatagramLwIp.cpp
@@ -26,6 +26,11 @@ namespace services
             else
                 return std::make_pair(Convert(std::get<Udpv6Socket>(socket).first), std::get<Udpv6Socket>(socket).second);
         }
+
+        u_int8_t DefaultNetifZone()
+        {
+            return netif_default != nullptr ? netif_default->ip6_addr->u_addr.ip6.zone : IP6_NO_ZONE;
+        }
     }
 
     DatagramExchangeLwIP::DatagramExchangeLwIP(DatagramExchangeObserver& observer)
@@ -83,7 +88,7 @@ namespace services
             IPv6Address ipv6Address = std::get<Udpv6Socket>(remote).first;
             ip_addr_t ipAddress IPADDR6_INIT(0, 0, 0, 0);
             IP6_ADDR(&ipAddress.u_addr.ip6, PP_HTONL(ipv6Address[1] + (static_cast<uint32_t>(ipv6Address[0]) << 16)), PP_HTONL(ipv6Address[3] + (static_cast<uint32_t>(ipv6Address[2]) << 16)), PP_HTONL(ipv6Address[5] + (static_cast<uint32_t>(ipv6Address[4]) << 16)), PP_HTONL(ipv6Address[7] + (static_cast<uint32_t>(ipv6Address[6]) << 16)));
-            ip6_addr_set_zone(&ipAddress.u_addr.ip6, netif_default->ip6_addr->u_addr.ip6.zone);
+            ip6_addr_set_zone(&ipAddress.u_addr.ip6, DefaultNetifZone());
             err_t result = udp_connect(control, &ipAddress, std::get<Udpv6Socket>(remote).second);
             assert(result == ERR_OK);
         }
@@ -113,7 +118,7 @@ namespace services
             IPv6Address ipv6Address = std::get<Udpv6Socket>(remote).first;
             ip_addr_t ipAddress IPADDR6_INIT(0, 0, 0, 0);
             IP6_ADDR(&ipAddress.u_addr.ip6, PP_HTONL(ipv6Address[1] + (static_cast<uint32_t>(ipv6Address[0]) << 16)), PP_HTONL(ipv6Address[3] + (static_cast<uint32_t>(ipv6Address[2]) << 16)), PP_HTONL(ipv6Address[5] + (static_cast<uint32_t>(ipv6Address[4]) << 16)), PP_HTONL(ipv6Address[7] + (static_cast<uint32_t>(ipv6Address[6]) << 16)));
-            ip6_addr_set_zone(&ipAddress.u_addr.ip6, netif_default->ip6_addr->u_addr.ip6.zone);
+            ip6_addr_set_zone(&ipAddress.u_addr.ip6, DefaultNetifZone());
             err_t result = udp_connect(control, &ipAddress, std::get<Udpv6Socket>(remote).second);
             assert(result == ERR_OK);
             result = udp_bind(control, IP6_ADDR_ANY, localPort);
@@ -276,7 +281,7 @@ namespace services
         {
             auto address = Convert(*remote).first;
             if (address.type == IPADDR_TYPE_V6)
-                ip6_addr_set_zone(&address.u_addr.ip6, netif_default->ip6_addr->u_addr.ip6.zone);
+                ip6_addr_set_zone(&address.u_addr.ip6, DefaultNetifZone());
 
             err_t result = udp_sendto(control, buffer, &address, Convert(*remote).second);
             assert(result == ERR_OK);
--- a/services/network/ConnectionMbedTls.cpp
+++ b/services/network/ConnectionMbedTls.cpp
@@ -13,7 +13,7 @@ extern "C"
     }
 #endif
 
-    int mbedtls_hardware_poll(void* data, unsigned char* output, size_t len, size_t* olen)
+    __weak int mbedtls_hardware_poll(void* data, unsigned char* output, size_t len, size_t* olen)
     {
         really_assert(services::MbedTlsAdapter::InstanceSet());
         services::MbedTlsAdapter::Instance().RandomDataGenerator().GenerateRandomData(infra::ByteRange(output, output + len));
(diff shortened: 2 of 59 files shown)


services/network/ConnectionMbedTls.cpp
```diff
@@ -1,19 +1,20 @@
...
+#include "mbedtls/platform_time.h"
+
     mbedtls_ms_time_t mbedtls_ms_time(void)
     {
         return static_cast<mbedtls_ms_time_t>(std::chrono::duration_cast<std::chrono::milliseconds>(infra::Now(3).time_since_epoch()).count());
     }
 #endif
 
-    int mbedtls_hardware_poll(void* data, unsigned char* output, size_t len, size_t* olen)
+    [[gnu::weak]] int mbedtls_hardware_poll(void* data, unsigned char* output, size_t len, size_t* olen)
```
richardapeters, line 17: weak symbols work very confusing in libraries. Can you find a different way to either include mbedtls_hardware_poll or not?

lwip/lwip_cpp/DatagramLwIp.cpp
```diff
@@ -26,6 +26,11 @@ namespace services
             else
                 return std::make_pair(Convert(std::get<Udpv6Socket>(socket).first), std::get<Udpv6Socket>(socket).second);
         }
+
+        u8_t DefaultNetifZone()
+        {
+            return netif_default != nullptr ? netif_default->ip6_addr->u_addr.ip6.zone : IP6_NO_ZONE;
```
richardapeters, line 32: I see a lot of code added to deal with the case that there is no netif_default. But what should our code do when there is no default interface?

I can think of a case when there are two interfaces but neither is default; in that case, our code should be instructed to select either interface explicitly.

Or there is one interface; in that case that interface can be the default.

Or there is no interface at all; in that case, why are we creating sockets?

fabiangottstein: Agreed, although handling multiple interface is currently not possible (in embedded-infra-lib) at all as I understand it?

services/network/ConnectionMbedTls.cpp
Petzep: The addition of this single attribute is a result of an elaborate research, hence my long explanation as it also helped restructuring my thoughts. I split my reply up into two parts for that matter.

## Motivation
`mbedTLS` provides a structured way to implement your own entropy generator. ([mbedTLS docs](https://mbed-tls.readthedocs.io/en/latest/kb/development/hw_acc_guidelines/#alternative-cryptography-engines-implementation))

https://github.com/philips-software/amp-embedded-infra-lib/blob/dfa56147333624cdfab31ea30045c45306798fb6/external/crypto/mbedtls/mbedtls_emil_config.h#L546-L548

The `embedded-infra-lib` uses this alternative implementation to link the entropy generation to `hal::SynchronousRandomDataGenerator` this ensures that `mbedTLS` uses the same method of random data generation as the rest of the library. In this way custom entropy generation is supported by the platform.

But, an issue arises when a vendor (e.g. Espressif) uses `MBEDTLS_ENTROPY_HARDWARE_ALT` to link their vendor specific entropy generator. [1]

[esp-idf/components/mbedtls/port/esp_hardware.c at master · espressif/esp-idf](https://github.com/espressif/esp-idf/blob/8d7d8aef58829a3afc0859ff76305decce45ed55/components/mbedtls/port/esp_hardware.c#L70-L76)
````
int mbedtls_hardware_poll(void *data, unsigned char *output, size_t len, size_t *olen)
{
    esp_fill_random(output, len);
    *olen = len;
    return 0;
}
````

This will cause a redefinition error during compilation.

My solution was making `embedded-infra-lib`'s function declaration weak to allow for vendor specific overrides. Because weak-linking is not a C++ language feature, I could not justify adding something like `EMIL_WEAK` to [infra/util/Compatibility.hpp](https://github.com/philips-software/amp-embedded-infra-lib/blob/main/infra/util/Compatibility.hpp). Therefore, I chose the most compatible solution with the least amount of changes.

I do agree, weak-linking symbols in libraries has a confusing effect.

## Alternative
Make the definition conditional on `EMIL_EXTERNAL_MBEDTLS_TARGET` to allow custom mbedTLS targets to include their own implementation.

## Future alternative
Having said that, the use of `MBEDTLS_ENTROPY_HARDWARE_ALT` is not recommended anymore since the release of `TF-PSA-Crypto` `1.0.0` 
[TF-PSA-Crypto/ChangeLog at development · Mbed-TLS/TF-PSA-Crypto](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/main/ChangeLog)
> The custom entropy collector callback mbedtls_hardware_poll() (enabled by MBEDTLS_ENTROPY_HARDWARE_ALT) has been replaced by a new callback mbedtls_platform_get_entropy() with different parameters (enabled by MBEDTLS_PSA_DRIVER_GET_ENTROPY). See the new function's documentation and “[Custom entropy collector](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/main/docs/1.0-migration-guide.md#custom-entropy-collector)” in the migration guide.

If, in a future effort, the `embedded-infra-lib` moves to a PSA implementation of mbedTLS, `mbedtls_psa_external_get_random()` is prefered instead of `mbedtls_hardware_poll()`. 
Following the [migration guide](https://github.com/Mbed-TLS/TF-PSA-Crypto/blob/main/docs/1.0-migration-guide.md#entropy-configuration)

### Footnotes
[1]: Espressif uses the same (_if-linux-build-then_)-reasoning to enable or disable the custom entropy functions. https://github.com/espressif/esp-idf/blob/master/components/mbedtls/esp_tee/esp_tee_mbedtls_config.h#L30-L35

File: lwip/lwip_cpp/DatagramLwIp.cpp (hunk 2 above)
Comment (Petzep) [reply lwip/lwip_cpp/DatagramLwIp.cpp]: Thanks for the feedback, it made me aware of the fail-fast principle in coding.
The additions will be reverted

File: CMakeLists.txt [hunk 3]
```diff
@@ -114,6 +114,14 @@ option(EMIL_INCLUDE_SEGGER_RTT "Include support for Segger RTT" Off)
 set(EMIL_EXTERNAL_LWIP_TARGET "" CACHE STRING "Specify an external LWIP target")
 option(EMIL_LWIP_CHECKSUM_IN_SOFTWARE "Enable lwIP checksum calculation in software" Off)
 
+if (EMIL_EXTERNAL_MBEDTLS_TARGET STREQUAL "")
+    set(EMIL_MBEDTLS_CUSTOM_ENTROPY_DEFAULT On)
+else()
+    set(EMIL_MBEDTLS_CUSTOM_ENTROPY_DEFAULT Off)
+endif()
+
+option(EMIL_MBEDTLS_CUSTOM_ENTROPY "Enable custom entropy for MbedTLS" ${EMIL_MBEDTLS_CUSTOM_ENTROPY_DEFAULT})
```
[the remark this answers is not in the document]
Petzep: Added an extra check to the cmake to prevent his configuration from happening.

services/network/CMakeLists.txt
```diff
@@ -115,6 +115,8 @@ if (EMIL_INCLUDE_MBEDTLS OR NOT EMIL_EXTERNAL_MBEDTLS_TARGET STREQUAL "")
         )
     endif()
 
+    target_compile_definitions(services.network PRIVATE EMIL_MBEDTLS_CUSTOM_ENTROPY=$<BOOL:${EMIL_MBEDTLS_CUSTOM_ENTROPY}>)
```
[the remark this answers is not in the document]
Petzep: Added an extra check to the cmake to prevent his configuration from happening.

PR #1340, opened 2026-09-14
fix(infra/syntax): report a JSON number that does not fit int64_t

CaptainAni187:
`JsonStreamingParser` accumulates a number literal directly into the `int64_t` it hands to visitors:

```cpp
tokenNumber = tokenNumber * 10 + c - '0';
```

Once the literal passes the end of that range the accumulation overflows. Signed overflow is undefined, and the value that came out of it was delivered to the visitor as a different, plausible-looking number — with the sign flipped in both directions:

| literal | `VisitNumber` received |
|---|---|
| `9223372036854775807` | `9223372036854775807` |
| `9223372036854775808` | `-9223372036854775808` |
| `99999999999999999999` | `7766279631452241919` |
| `-9223372036854775809` | `9223372036854775807` |

Nothing was reported to the visitor in the three wrong cases; the parser carried on and called `Close()`. A device reading a length, an index or a threshold out of untrusted JSON acts on the wrong number with no indication anything went wrong.

UBSan on the current `main` confirms it is undefined behaviour rather than benign wraparound:

```
JsonStreamingParser.cpp:99:52: runtime error: signed integer overflow:
9223372036854775800 + 55 cannot be represented in type 'int64_t'
```

## The change

Scan the magnitude in a `uint64_t` with the sign kept apart in `tokenSign`, which is how `JsonTokenizer` already splits a number in `Json.cpp` (`ToInt<uint64_t>` plus `JsonBiggerInt(value, negative)`). Unsigned arithmetic makes the accumulation itself well-defined, and each digit is range-checked against what the sign allows — a negative literal reaches one further from zero than a positive one.

A literal that does not fit becomes `Token::numberOverflow`. Both the object and the array parser handle it the way they handle `Token::stringOverflow`: the value is closed and the visitor gets a new `NumberOverflow()` callback, then parsing carries on with the rest of the object or array. `NumberOverflow()` sits next to `StringOverflow()` on `JsonVisitor`, with empty defaults on `JsonObjectVisitor` and `JsonArrayVisitor`, a forward on `JsonObjectVisitorDecorator`, and a mock method.

`SignedTokenNumber()` converts magnitude and sign back to `int64_t` without negating a value the type cannot hold, so `-9223372036854775808` still arrives intact.

Numbers inside a sub-object the visitor declined to parse are still ignored rather than turned into an error, unchanged from before.

The fractional and exponent digits are still discarded, so `5.123e456` continues to arrive as `5`. That is pre-existing behaviour and I have left it alone rather than widen this PR.

## Tests

Five added to `TestJsonStreamingParser.cpp` fail without the change:

- `number_above_int64_t_results_in_NumberOverflow`
- `number_below_int64_t_results_in_NumberOverflow`
- `number_of_more_digits_than_int64_t_holds_results_in_NumberOverflow`
- `parsing_continues_after_NumberOverflow`
- `JsonStreamingObjectParserArrayTest.number_above_int64_t_results_in_NumberOverflow`

`VisitNumber_with_the_extremes_of_int64_t` passes on `main` too; it pins the boundary so the range check can't start rejecting values that do fit.

The overflow was found by fuzzing `JsonStreamingParser` under ASan and UBSan with the input split into varying chunk sizes. The checked-in `FuzzJson.cpp` covers `infra::JsonObject` only, so the streaming parser has no fuzz target today; I am happy to open a separate PR adding one if that would be useful.


richardapeters (changes requested):
Overall I like the change, but if overflow is now detected then one thing is missing: Similar to how StringOverflow() is invoked when a string is found that does not fit in the buffer, there should be a NumberOverflow() added to JsonObjectVisitor, which is invoked on any number overflow

CaptainAni187:
Makes sense. Added NumberOverflow() next to StringOverflow() (defaults on both visitors, forwarded by the decorator, added to the mocks), and an out-of-range number now goes there instead of ParseError, with parsing continuing afterwards like it does for strings. Tests are switched over, plus one checking the rest of the object still gets parsed. Also rebased on main.

